PR #3, opened 2026-04-13
Handle comparisons against translated functions

lucic71:
This PR adds support for translating comparisons against functions that have translation rules `fn_ptr == translated_rule`, for example in `fn1 == fread` in fn_ptr_stdlib_compare.cpp:

```rs
fn1 == 
  FnPtr::new<fn(AnyPtr, u64, u64, Ptr<::std::fs::File>) -> u64>(
    libcc2rs::fread_refcount)
```


lucic71:
This is still a draft because not all unit tests pass:
- [x] function pointer casting
- [x] comparing local function pointer with stdlib function pointer. This is especially tricky for stdlib function pointers that have associated translation rules

Edit: these issues are fixed

lucic71:
Will split this into multiple PRs:
* translate function pointers and non-capturing lambdas as Option<fn> in unsafe and FnPtr in refcount
* expose translation rules function declarations to generated code to allow `fn_ptr == translated_rule`
* minor changes described here https://github.com/Cpp2Rust/cpp2rust/pull/3#issue-4257709400

lucic71:

rules/stdio/tgt_refcount.rs
```diff
@@ -55,7 +55,7 @@ fn f4(a0: &mut ::std::fs::File, a1: i64, a2: i32) -> i32 {
     }
 }
 
-fn f5(a0: AnyPtr, a1: u64, a2: u64, a3: Ptr<::std::fs::File>) -> u64 {
+pub fn f5(a0: AnyPtr, a1: u64, a2: u64, a3: Ptr<::std::fs::File>) -> u64 {
```
nunoplopes: why these changes?

cpp2rust/converter/translation_rule.h
```diff
@@ -61,6 +61,8 @@ struct TypeInfo {
 };
 
 struct ExprTgt {
+  std::string name;
```
nunoplopes: what are these things and what are they used for? We can't keep growing these. We should shrink them to speedup, not grow.

lucic71: They are used to take the address of a function that has a translation rule, for example `fread` becomes `rules::{name}_{module}::{name_of_fread_rule}` -> `rules::stdio_tgt_unsafe::f5`. All this happens in GetRulePathForFunction

rules/stdio/tgt_refcount.rs
lucic71: To make f5 visible outside the crate so that translted code can take the address of this function: `rules::stdio_tgt_refcount::f5`

nunoplopes: That's a major hack. No way this works. Rule's names are internal and should not be exported.
Translated programs only link with libcc2rs. Add new symbols there.

lucic71: So generated code should use `libcc2rs::fread` instead of `rules::stdio_tgt_refcount::f5`? In this case `libcc2rs::fread` should be defined as a call to `rules::stdio_tgt_refcount::f5`. For this to work, I think making f5 public is still required

cpp2rust/converter/converter_lib.cpp
```diff
@@ -660,4 +662,18 @@ clang::Expr *CreateConversionToBool(clang::Expr *expr, clang::ASTContext &ctx) {
       /*BasePath=*/nullptr, clang::VK_PRValue, clang::FPOptionsOverride());
 }
 
+static void Trim(std::string &s) {
+  auto is_space = [](unsigned char c) { return std::isspace(c); };
+  s.erase(s.begin(), std::find_if_not(s.begin(), s.end(), is_space));
+  s.erase(std::find_if_not(s.rbegin(), s.rend(), is_space).base(), s.end());
```
nunoplopes: instead of modifying the string, it's cheaper to do substr on a string_view

cpp2rust/converter/converter_lib.cpp
```diff
@@ -660,4 +662,18 @@ clang::Expr *CreateConversionToBool(clang::Expr *expr, clang::ASTContext &ctx) {
...
+}
+
+void Unwrap(std::string &s, std::string_view prefix, std::string_view suffix) {
+  Trim(s);
+  if (s.size() >= prefix.size() + suffix.size() && s.starts_with(prefix) &&
```
nunoplopes: this size comparison is not needed, unless prefix and suffix might be the same?

rules/stdio/tgt_refcount.rs
nunoplopes: No, we should not export anything in rules. And should definitely not export a function named f5.
You need to create an fread function in libcc2rs.

lucic71: We also need to support calling into mapped functions using function pointers, like:

```cpp
auto fn = &fread;
fn(...);
```

This means that the body of `libcc2rs::fread` and `rules::stdio_tgt_refcount::f5` has to be exactly the same. Without exporting f5 from rules, we will copy-paste the body of `rules::stdio_tgt_refcount::f5` inside `libcc2rs::fread`

nunoplopes: That's ok for now. But we are not exporting the rules directory and f5 names.

lucic71: I created `fread_refcount` and `fread_unsafe`. Generated code can use these functions for saving the address of fread. Generated code can also call into fread through a function pointer.

To avoid code duplication, the fread translation rules also call into `fread_refcount` and `fread_unsafe`

cpp2rust/converter/converter_lib.cpp
lucic71: I moved this check in an assert so that it only triggers in Debug builds

cpp2rust/converter/converter_lib.cpp
lucic71: I moved to string_view

PR #4, opened 2026-04-16
ErasedPtr: drop Clone bound and allow cross-type nullptr casts

lucic71:
This PR allows converting between `void*` and non-clonable pointer types (`FILE*`, for example), and fixes cross-type nullptr casts (casting a `nullptr void*` to any other pointer).

This is a smaller PR based of #3


lucic71:
To implement the second part of #3, i.e. "expose translation rules function declarations to generated code to allow fn_ptr == translated_rule", this PR and #7 need to be merged

PR #6, opened 2026-04-16
Speed-up unit tests

lucic71:
Sub-part of #3

First unit test compiles libcc2rs and rules in cargo-target, then all unit-tests use the precompiled libraries from cargo-target instead of building them each time.


nunoplopes:
Isn't this missing the removal of Cargo.toml, or is that part of another PR?

lucic71:
> Isn't this missing the removal of Cargo.toml, or is that part of another PR?

I removed it now, thanks!

tests/lit/lit/formats/Cpp2RustTest.py
```diff
@@ -7,21 +7,29 @@
 import difflib, os, re, shutil, random
 import tomli
 
-def read_rust_version():
```
nunoplopes, line 10: this change (turning into a method) is not useful; it makes the code slower for no benefit.

lucic71: Should I also make sharedTargetDir and cargoEnv free functions instead of methods?

nunoplopes: yep

lucic71: Done

PR #7, opened 2026-04-17
Translate function pointers and lambda

lucic71:
Sub-part of #3, only covers: translate function pointers and non-capturing lambdas as Option<fn> in unsafe and FnPtr in refcount

* deleted the `fn_ptr!` macro, address is computed inside `FnPtr::new` using the `FnAddr` trait
* replaced `cast_history` with the original function pointer + current cast
* in unsafe function pointers + casts are translated using: `Option<fn>` + `std::mem::transmute`
* in refcount function pointers + casts are translated using: `FnPtr<fn>` + `FnPtr<fn>::cast`

### DIFF
--- a/libcc2rs/src/fn_ptr.rs
+++ b/libcc2rs/src/fn_ptr.rs
@@ -31,15 +31,19 @@ macro_rules! impl_fn_addr {
 }
 impl_fn_addr!();
 
-#[derive(Clone)]
-pub(crate) struct FnState {
-    addr: usize,
-    original: Rc<dyn Any>,
-    current_cast: Option<Rc<dyn Any>>,
+trait ErasedFn: Any {
+    fn addr(&self) -> usize;
+}
+
+impl<T: FnAddr + Any> ErasedFn for T {
+    fn addr(&self) -> usize {
+        self.fn_addr()
+    }
 }
 
 pub struct FnPtr<T> {
-    state: Option<Rc<FnState>>,
+    original: Option<Rc<dyn ErasedFn>>,
+    current_cast: Option<Rc<dyn ErasedFn>>,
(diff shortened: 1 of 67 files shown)


libcc2rs/src/fn_ptr.rs
```diff
@@ -0,0 +1,179 @@
...
+
+#[derive(Clone)]
+pub(crate) struct FnState {
+    addr: usize,
+    original: Rc<dyn Any>,
+    current_cast: Option<Rc<dyn Any>>,
+}
+
+pub struct FnPtr<T> {
+    state: Option<Rc<FnState>>,
```
nunoplopes: can't we embed the 3 fields here?
if null -> addr = 0.
Otherwise, the 3 fields are set. I don't see the need for this indirection.

libcc2rs/src/fn_ptr.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    #[inline]
+    pub fn is_null(&self) -> bool {
+        self.state.is_none()
+    }
+}
+
+impl<T: FnAddr + 'static> FnPtr<T> {
+    pub fn new(f: T) -> Self {
+        let addr = f.fn_addr();
```
nunoplopes: can't this be computed dynamically from FnState?

lucic71: I added a new trait ErasedFn that has an `addr` method which computes the address dynamically

libcc2rs/src/fn_ptr.rs
lucic71: I dropped the `addr` member ,`addr` is computed dynamically using the `ErasedFn` trait.

I inlined original and current_cast in FnPtr

PR #12, opened 2026-04-19
Translate C boolean expressions

lucic71:
In C, as opposed to C++, clang does not generate IntegerToBoolean implicit casts. To fix this, create an artificial implicit cast so that C and C++ code can be handled uniformly.

As a side effect of this PR, C unit tests are now translated in C instead of C++. va_arg_struct_ctx.c does not translate well in C yet, this will be handled in #10


lucic71:
This fails on macOS, I will do another PR to fix it

PR #20, opened 2026-04-20
Fix detection of va_list type

lucic71:
On macOS `va_list` is typedef'ed to `char*`, on Linux is typedefed to `__va_list_tag[1]`. Traverse the TypedefType/AdjustedType in order to check if a type is va_list instead of checking against the Linux specific `__va_list_tag`.

### DIFF
--- a/cpp2rust/converter/converter.cpp
+++ b/cpp2rust/converter/converter.cpp
@@ -63,15 +63,15 @@ bool Converter::VisitRecoveryExpr(clang::RecoveryExpr *expr) {
 }
 
 bool Converter::Convert(clang::QualType qual_type) {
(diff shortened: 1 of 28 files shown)


lucic71:
Fixes the macOS errors in #12

lucic71:
In C++ mode, for va_list, on both macOS and Ubuntu, clang typedefs to `struct __va_list_tag[1]`.

In C mode, clang typedefs to `struct __va_list_tag[1]` on Ubuntu and to `char*` on macOS.

That's why the va_arg tests broke after switching to C mode.

nunoplopes:
3 tests still failing

lucic71:
> 3 tests still failing

It works now

PR #23, opened 2026-04-21
Handle more sophisticated switch/case patterns

lucic71:


lucic71:
Will split this into 3 PRs:
* default cases in the middle
* case chains containing default
* fallthrough

PR #28, opened 2026-04-23
Add switch tests

lucic71:
This adds all the switch/case tests introduced in #23

Not all tests compile, but that's fine, they will be fixed in future PR's. Some tests are marked with the new

```cpp
// translation-fail
```

marker that's aded to tests which break the translator

### DIFF
--- a/tests/lit/lit/formats/Cpp2RustTest.py
+++ b/tests/lit/lit/formats/Cpp2RustTest.py
@@ -96,10 +96,6 @@ class Cpp2RustTest(TestFormat):
       shutil.rmtree(tmp_dir, True)
       return code, str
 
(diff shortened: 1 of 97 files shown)


tests/lit/lit/formats/Cpp2RustTest.py
```diff
@@ -113,8 +111,14 @@ def fail(str, code = fail_code):
       generated = f.read()
 
     if returncode != 0:
+      if should_not_translate:
+        return lit.Test.PASS, ''
```
nunoplopes: change this to XFAIL so we can track these

lucic71: Done

PR #30, opened 2026-04-27
Translate switch with default case in the middle

lucic71:
This PR handles code as:

```cpp
// tests/unit/switch_default_middle.cpp
int default_middle(int x) {
  int r = 0;
  switch (x) {
  case 1:
    r = 1;
    break;
  default:
    r = 99;
    break;
  case 2:
    r = 2;
    break;
  }
  return r;
}
```

In rust, the default arm has to be the last arm, otherwise, arms following the default one will be ignored, making this assertion fail `assert(default_middle(2) == 2)`. Hence, the rust translation becomes:

```rs
 match x {
     v if v == 1 => {
         r = 1;
     }
     v if v == 2 => {
         r = 2;
     }
     _ => {
         r = 99;
     }
 }
```

This is achieved by iterating over all top-level case statements in VisitSwitchStmt and deferring the translation of the default arm until every othter arm is translated. A top-level case statement is defined as:

```cpp
switch (x) {
case TOP_LEVEL_STMT_1:
case NOT_TOP_LEVEL_STMT_1:
default: // not top level as well
  ...
  break;
case TOP_LEVEL_STMT_2:
case NOT_TOP_LEVEL_STMT_2:
  ...
  break
}
```

Furthermore, chains of case statements that contain a default statement are reduced to default, effectively making the above code translate as:

```rs
match x {
  v if v == TOP_LEVEL_STMT_2 || v == NOT_TOP_LEVEL_STMT_2 => {}
  _ => {} // TOP_LEVEL_STMT_1, NOT_TOP_LEVEL_STMT_1, default was reduced to default
}
```


cpp2rust/converter/converter.cpp
```diff
@@ -2621,63 +2621,67 @@ bool Converter::VisitImplicitValueInitExpr(clang::ImplicitValueInitExpr *expr) {
...
+                              bool is_default) {
+  if (is_default) {
+    StrCat("_ => {");
   } else {
-    if (clang::isa<clang::CaseStmt>(stmt)) {
-      StrCat(" => {");
-    } else {
-      StrCat("_ => {");
-    }
+    StrCat("v if v == ");
```
nunoplopes, line 2654: instead of `v == 1 || v == 2 || v == 3`, can't you do:
```rust
 match x {
            1 | 2 | 3 => ...
```

lucic71: It's because of enums. A switch such as:

```cpp
enum E {
        A,
        B,
        C
};

int main() {
        enum E e = A;
        switch (e) {
        case A:
        case B:
                break;
        case C:
                break;
        }
        return 0;
}
```

would get translated as:

```rs
        match e {
            // (E::A as u32)  is an expression, not a pattern, so it's not allowed to be chained using |
            (E::A as u32) | (E::B as u32) => {
                break;
            }
            v if v == (E::C as u32) => {
                break;
            }
            _ => {}
        }
```

I think we can just ignore the implicit enum to integer cast in order to allow this to work with | instead of ||

nunoplopes: Ok, then let's keep it as-is. In C++ you can mix enums and integers.

nunoplopes:
Please rebase; the branch has conflicts.

lucic71:
> Please rebase; the branch has conflicts.

Done

nunoplopes:
CI failed.

lucic71:
> CI failed.

Fixed

PR #32, opened 2026-04-27
Translate unions in unsafe

lucic71:
C/C++ unions are translated as Rust unions in unsafe. They manually implement default with `std::mem::zeroed` since unions cannot derive Default

Each struct and union is decorated with the `repr(C)` attribute to make sure that the C layout is preserved. We need this because Rust is free to reorder fields for optimal struct layout


cpp2rust/converter/converter.cpp
```diff
@@ -3211,20 +3233,26 @@ void Converter::AddDefaultTrait(const clang::CXXRecordDecl *decl) {
...
-        /*ZeroInitialization=*/false, clang::CXXConstructionKind::Complete,
-        clang::SourceRange()));
-  } else {
-    StrCat(struct_name);
+
+  if (auto *cxx = clang::dyn_cast<clang::CXXRecordDecl>(decl)) {
+    if (auto *default_ctor = GetUserDefinedDefaultConstructor(cxx)) {
+      StrCat(keyword_unsafe_);
+      PushBrace unsafe_brace(*this);
+      Convert(clang::CXXConstructExpr::Create(
```
nunoplopes, line 3241: won't this leak memory?

lucic71: It's created inside the allocator of ASTContext, the node is automatically freed when ASTContext is destroyed

PR #36, opened 2026-04-28
Translate string literals in C and C++

lucic71:
This PR adds support for declaring and using const/non-const pointers and arrays containing string literals. VisitStringLiteral now always generates a string literal (`[u8; N]` in unsafe, `Box<[u8]>` in refcount). It's the job of `VisitImplicitCast::CK_ArrayToPointerDecay` to convert the string literal to poiner if it's necessary.

In C string literals are `char[]` and in C++ they are `const char[]`. `VisitImplicitCast::CK_NoOp` handles the conversion between const and non-const char pointers.

Besides always generating a string literal instead of pointer, VisitStringLiteral also pad with null bytes the following code: `char array_bigger_than_string_literal[10] = "1"`, effectively becoming `"1\0\0\0..."`.


tests/unit/string_literals.cpp
```diff
@@ -0,0 +1,35 @@
+void foo_mut(char *str) {}
+void foo_const(const char *str) {}
+
+int main() {
+  char *mutable_strings[] = {"a", "b", "c"};
```
nunoplopes: this code is invalid in C++. strings are const.

tests/unit/out/refcount/string_literals_sized.rs
```diff
@@ -0,0 +1,39 @@
...
+use std::collections::BTreeMap;
+use std::io::prelude::*;
+use std::io::{Read, Seek, Write};
+use std::os::fd::AsFd;
+use std::rc::{Rc, Weak};
+pub fn main() {
+    std::process::exit(main_0());
+}
+fn main_0() -> i32 {
+    let  empty_buf : Value<Box<[ u8  ]> > = Rc::new(RefCell::new(Box::<[u8]>::from(b"\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0".as_slice()) )) ;
```
nunoplopes: this is a bit ugly..
Can we detect this case and emit `vec![0; 256].into_boxed_slice()`?

tests/unit/string_literals.cpp
lucic71: Both GCC and Clang generate a warning in this case. Should we reject this code or accept it?

nunoplopes: I would just remove these tests.

tests/unit/out/refcount/string_literals_sized.rs
lucic71: Done

tests/unit/string_literals.cpp
lucic71: Done

PR #47, opened 2026-04-30
Translate void casts

lucic71:


cpp2rust/converter/converter.cpp
```diff
@@ -1865,10 +1865,21 @@ bool Converter::VisitImplicitCastExpr(clang::ImplicitCastExpr *expr) {
   return false;
 }
 
+void Converter::ConvertVoidCastExpr(clang::ExplicitCastExpr *expr) {
+  PushExprKind push(*this, ExprKind::Void);
+  StrCat("let _ = ");
```
nunoplopes: I don't think that the let is needed. You can just evaluate the expression as a statement. No clone needed either.

lucic71: Done. I used `let _ = expr.clone()` to avoid moving expr, but simply writing `expr;` is enough

lucic71: Actually no, the move still ocurs if expr is not Copy, for example:

```cpp
int *p;
(void) p;
```

So a clone is needed for those types

lucic71: Fixed. I now add clone for non-Copy types

PR #50, opened 2026-05-01
Disable verbose logging by default

lucic71:
Logging can be re-enabled using the `--verbose` option

### DIFF
--- a/cpp2rust/converter/models/converter_refcount.cpp
+++ b/cpp2rust/converter/models/converter_refcount.cpp
@@ -824,7 +824,7 @@ static std::vector<const char *> printf2fmt(std::string &format) {
         }
       }
     }
-    cpp2rust::verrs() << "Unknown printf format: " << format << "\n";
+    verrs() << "Unknown printf format: " << format << "\n";
     assert(0);
   }
   return types;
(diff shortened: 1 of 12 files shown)


cpp2rust/converter/models/converter_refcount.cpp
```diff
@@ -823,7 +824,7 @@ static std::vector<const char *> printf2fmt(std::string &format) {
         }
       }
     }
-    llvm::errs() << "Unknown printf format: " << format << "\n";
+    verrs() << "Unknown printf format: " << format << "\n";
```
nunoplopes: These kinds of things must be printed always. This is not about being verbose or not

cpp2rust/converter/models/converter_refcount.cpp
```diff
@@ -850,7 +851,7 @@ void ConverterRefCount::ConvertPrintf(clang::CallExpr *expr) {
   } else if (fd == "stderr" || fd == "__stderrp") {
     StrCat(ends_newline ? "eprintln!(" : "eprint!(");
   } else {
-    llvm::errs() << "Uknown fprintf fd: " << fd << "\n";
+    verrs() << "Uknown fprintf fd: " << fd << "\n";
```
nunoplopes: same here

cpp2rust/logging.h
```diff
@@ -0,0 +1,15 @@
...
+#pragma once
+
+#include <llvm/Support/raw_ostream.h>
+
+namespace cpp2rust {
+
+void SetVerbose(bool verbose);
+
+llvm::raw_ostream &verrs();
+llvm::raw_ostream &vouts();
```
nunoplopes: we probably only need one stream for verbose

cpp2rust/converter/models/converter_refcount.cpp
lucic71: Done

cpp2rust/converter/models/converter_refcount.cpp
lucic71: Done

cpp2rust/logging.h
lucic71: I removed vouts and renamed verrs to log

PR #61, opened 2026-05-04
Parse src.cpp rules at build time

### LINKED ISSUE #55: Parsing rules's source (C++) is the bottleneck
Cpp2Rust is getting slower when we add new rules. To load the rule set, we load preprocessed json corresponding to Rust code (quick), and parse the source C++ code using the clang library. This process is very slow, as it needs to parse many C++ headers besides our code.

We need to preprocess the C++ code as well, but the result cannot be checked into git because C++ libraries differ in naming. Must be done at compile time.

lucic71:
Fixes #55 

Each src.cpp from the `rules/` dir is now parsed at build time, instead of on every invocation of cpp2rust. This gives a 5-8x speedup in running the unit tests in the Github runner. Translating a simple unit test like `assert.cpp` went down form 6 seconds to 0.1 seconds.

The new `cpp-rule-preprocessor` binary from `cpp2rust/` is responsible for parsing one `src.cpp` at a time and to generate a `ir_src.json` in the same dir as the `src.cpp` file. The format of the json file is: `f<n> / t<n>: { "to_string": <Mapper::ToString> }`.

All `src.cpp` files are parsed in parallel, CMakeLists takes care of that. The `check-rules` target is now responsible for parsing the Rust rules and the C++ rules. The IR for `src.cpp` is not indexed in git. The IR for the Rust rules remains unchanged.

I ported the clang AST matcher form translation_rule.cpp into the new file cpp_rule_preprocessor.cpp. As a result, translation_rule.cpp is only responsible for loading IR JSONs.


.github/workflows/run-tests.yml
```diff
@@ -58,14 +58,14 @@ jobs:
...
 
-      - name: Run unit tests
-        run: ninja check
-        working-directory: build
-
       - name: Check rules
         run: ninja check-rules
         working-directory: build
 
+      - name: Run unit tests
```
nunoplopes, line 65: this change is weird. `ninja` should regen the files.

CMakeLists.txt
```diff
@@ -135,20 +135,41 @@ add_custom_target("check"
...
+  add_custom_command(
+    OUTPUT ${_out}
+    COMMAND $<TARGET_FILE:cpp-rule-preprocessor> --file ${_src}
+    DEPENDS ${_src} cpp-rule-preprocessor
+    VERBATIM
+  )
+  list(APPEND cpp_rules_ir_outputs ${_out})
+endforeach()
+
+add_custom_target("preprocess-cpp-rules" DEPENDS ${cpp_rules_ir_outputs})
```
nunoplopes: add ALL so it runs with the build

CMakeLists.txt
```diff
@@ -135,20 +135,41 @@ add_custom_target("check"
...
+
+add_custom_target("preprocess-rust-rules"
```
nunoplopes: this needs to be changed to use the same pattern as above, so it runs on build and iff some rust file changed.

CMakeLists.txt
```diff
@@ -135,20 +135,41 @@ add_custom_target("check"
...
                               CARGO_TARGET_DIR=${PROJECT_SOURCE_DIR}/rule-preprocessor/target
                               cargo run
-                  COMMAND ${CMAKE_COMMAND} -E chdir ${PROJECT_SOURCE_DIR}
-                          git diff --exit-code -- rules/
                   WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}/rules"
                   DEPENDS install-rust-toolchain
                   USES_TERMINAL
                  )
 
+add_custom_target("check-rules"
```
nunoplopes, line 182: move the delete here and call preprocess & then diff.

.github/workflows/run-tests.yml
lucic71: Right, I will make the change to regen the files on `ninja`

However, I think it's better to run `check-rules` before unit tests because it's cheaper and catches the rules errors early

CMakeLists.txt
lucic71: Done. check-rules now contains deltee + call preprocess rust/cpp + diff

CMakeLists.txt
lucic71: Done. preprocess-rust-rules now has a list on dependent files and runs on all + when a file changes

CMakeLists.txt
lucic71: Done. preprocess-cpp-rules has ALL

.github/workflows/run-tests.yml
lucic71: Now that we run preprocess-{rust,cpp}-rules in ALL, check-rules does not need to delete then re-run preprocess-{rust,cpp}-rules, it can only do git diff

CMakeLists.txt
```diff
@@ -135,18 +135,57 @@ add_custom_target("check"
...
+          ${CMAKE_COMMAND} -E env
+              CARGO_TARGET_DIR=${PROJECT_SOURCE_DIR}/rule-preprocessor/target
+              cargo run
+  WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}/rules"
+  DEPENDS ${rust_rules_inputs} ${rule_preprocessor_sources}
+  VERBATIM
+)
+
+add_custom_target("preprocess-rust-rules" ALL
+                  DEPENDS ${rust_rules_ir_outputs} install-rust-toolchain)
```
nunoplopes, line 180: doesn't seem like it needs the rust toolchain here.

PR #68, opened 2026-05-11
Setup rust for C++ code analyzer

lucic71:
`ninja all` preprocesses the rules which requires the Rust environment to be set.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -35,7 +35,6 @@ jobs:
         llvm-version: 22
 
     - name: Setup Rust
-      if: matrix.language == 'rust'
       uses: dtolnay/rust-toolchain@master
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -35,7 +35,6 @@ jobs:
         llvm-version: 22
 
     - name: Setup Rust
-      if: matrix.language == 'rust'
```
nunoplopes, line 38: not needed for actions, just cpp and rust.

lucic71: I'm not sure I understand what you're suggesting

nunoplopes: != actions

lucic71: Done

PR #83, opened 2026-05-12
Translate function pointers passed in variadic arguments

lucic71:
In unsafe, the function pointer is first converted to void* when passed to a variadic function. Then, inside the variadic function, when calling `va_arg(ap, fn_ptr)`, the function is first retrieved as void*, then transmuted to function pointer. The `function pointer -calling variadic function-> void*, void* -va_arg(ap, fn_ptr)-> transmute<function pointer>` is needed so that we don't add the unsafe transmute in libcc2rs.

In refcount, this dance is not needed. VaArg and VaArgGet are implemented for FnPtr safely.

### DIFF
--- a/cpp2rust/converter/converter.cpp
+++ b/cpp2rust/converter/converter.cpp
@@ -1361,6 +1361,16 @@ std::optional<std::string> Converter::TryPluginConvert(clang::CallExpr *call) {
   return std::nullopt;
 }
 
+void Converter::ConvertVariadicArg(clang::Expr *arg) {
+  if (arg->getType()->isFunctionPointerType()) {
+    PushParen p(*this);
+    Convert(arg);
+    StrCat(".map_or(::std::ptr::null_mut(), |f| f as *mut ::libc::c_void)");
+    return;
+  }
+  Convert(arg);
+}
+
 void Converter::ConvertVAArgCall(clang::CallExpr *expr) {
   if (IsBuiltinVaStart(expr)) {
     StrCat(ToString(expr->getArg(0)->IgnoreImpCasts()),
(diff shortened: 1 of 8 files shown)


cpp2rust/converter/converter.cpp
```diff
@@ -2603,6 +2613,18 @@ bool Converter::VisitVAArgExpr(clang::VAArgExpr *expr) {
   if (auto *cast = clang::dyn_cast<clang::ImplicitCastExpr>(va_list_expr)) {
     va_list_expr = cast->getSubExpr();
   }
+  if (expr->getType()->isFunctionPointerType()) {
+    StrCat("std::mem::transmute::<*mut ::libc::c_void", token::kComma);
+    Convert(expr->getType());
+    StrCat(">");
```
nunoplopes: '>'

lucic71: Done

tests/unit/out/unsafe/va_arg_fn_ptr.rs
```diff
@@ -0,0 +1,93 @@
...
+pub unsafe fn negate_1(mut x: i32) -> i32 {
+    return -x;
+}
+pub unsafe fn add_2(mut a: i32, mut b: i32) -> i32 {
+    return ((a) + (b));
+}
+pub unsafe fn apply_unary_3(mut x: i32, args: &[VaArg]) -> i32 {
+    let mut ap: VaList = VaList::default();
+    ap = VaList::new(args);
+    let mut fn_: Option<unsafe fn(i32) -> i32> = std::mem::transmute::<
```
nunoplopes, line 21: Why is Option needed here?

lucic71: We translate functions pointers to `Option<fn>` in unsafe. Option is needed to encode nullability

PR #92, opened 2026-05-13
Support src.cpp and src.c in the same rules dir

lucic71:
For C string functions, C++ defines const overloads. For example, in C, strchr is defined as:

```c
char *strchr(const char *s, int c);
```

In C++, it's overloaded as:

```cpp
const char *strchr(const char *s, int c);
char *strchr(char *s, int c);
```

Writing `char *f5(const char *a0, int a1) { return strchr(a0, a1); }` would result in the following error:

```
error: cannot initialize return object of type 'char *' with an rvalue of type 'const char *'
    6 | char *f6(const char *a0, int a1) { return strchr(a0, a1); }
```

For this reason, parse both `src.cpp` and `src.c` in a rules dir. `src.c` keeps C specific rules and `src.cpp` keeps C++ specific rules. The function names in the 2 functions cannot collide, cpp-rule-preprocessor checks against colliding names.


cpp2rust/cpp_rule_preprocessor.cpp
```diff
@@ -751,24 +751,40 @@ namespace {
...
+  for (const char *name : {"src.c", "src.cpp"}) {
+    auto path = dir / name;
+    if (!fs::exists(path)) {
+      continue;
+    }
+    llvm::errs() << "Preprocessing " << path.string() << '\n';
+    llvm::json::Object file_root;
+    cpp2rust::Extract(path, file_root);
+    for (auto &[k, v] : file_root) {
+      if (root.find(k) != root.end()) {
```
nunoplopes: !root.try_emplace(k, v).second

lucic71: Done

PR #95, opened 2026-05-13
Add fwrite in libcc2rs to allow taking its address

lucic71:


libcc2rs/src/io.rs
```diff
@@ -121,6 +121,104 @@ pub fn fread_refcount(a0: AnyPtr, a1: u64, a2: u64, a3: Ptr<::std::fs::File>) ->
...
+    let mut written_bytes: usize = 0;
+    let mut buffer: [u8; 8192] = [0; 8192];
+
+    while written_bytes < total {
+        let remaining = total - written_bytes;
+        let to_fill = std::cmp::min(buffer.len(), remaining);
+
+        for b in buffer.iter_mut().take(to_fill) {
+            *b = src.read();
+            src = src.offset(1);
```
nunoplopes: src += 1

lucic71: Done

libcc2rs/src/io.rs
```diff
@@ -121,6 +121,104 @@ pub fn fread_refcount(a0: AnyPtr, a1: u64, a2: u64, a3: Ptr<::std::fs::File>) ->
...
+        }
+
+        if off == 0 {
+            break;
+        }
+
+        written_bytes += off;
+    }
+
+    std::io::Write::flush(&mut writer).expect("flush failed");
```
nunoplopes: I don't think fwrite flushes the stream

lucic71: I deleted the flush

PR #101, opened 2026-05-15
Assert that both C++ and Rust rules use the same generics

lucic71:


cpp2rust/converter/translation_rule.cpp
```diff
@@ -299,6 +300,26 @@ void ExprRule::dump() const {
...
 
+void ExprRule::validate(const std::string &name) const {
+  if (src.empty()) {
+    llvm::errs() << name << '\n';
+    dump();
+    assert(0 && "Expr rule loaded from IR but has no src");
+  }
+
+  // Check that both source and target use the same generics.
+  for (unsigned i = 0; i < generics.size(); ++i) {
```
nunoplopes: i = 0, e = generics.size(); i != e

cpp2rust/converter/translation_rule.cpp
```diff
@@ -299,6 +300,26 @@ void ExprRule::dump() const {
...
+    auto placeholder = std::format("T{}", i + 1);
```
nunoplopes: instead of allocating memory on every generic for every rule, an alternative is to declare a global constant:
char generic_types[][] = ["T1", "T2", "T3", "T4"];

cpp2rust/converter/translation_rule.cpp
```diff
@@ -16,6 +16,21 @@ namespace cpp2rust::TranslationRule {
 
 namespace {
 
+struct GenericPlaceholders {
+  char storage[kMaxGenerics][3]{};
+
+  constexpr GenericPlaceholders() {
+    for (unsigned i = 0; i < kMaxGenerics; ++i) {
```
nunoplopes: I meant a constant array, not this..

lucic71: It's a constexpr array with kMaxGeneric elements so that we don't get hard to debug errors when kMaxGeneric changes

nunoplopes: You can use an std::array if you want to make it safe. asserts aren't that need as it throws if you access OOB.

lucic71: Ok, I can use std::array, but the idea was not to manually write {"T1", "T2", "T3", ...}, but instead be generated automatically based on kMaxGenerics

lucic71: Done, I replaced with std::array

PR #112, opened 2026-05-17
Suppress clone in iterator conversion

lucic71:
From, tests/unit/redundant_copy_in_conversion:

```cpp
  // Comparing const_iterator with iterator forces an implicit conversion of
  // `end` to const_iterator. The AST shape differs between platforms:
  //
  //   Linux: const_it == ConvertingCtor(end)
  //   macOS: const_it == ConvertingCtor(CopyCtor(end))
  //
  // The extra inner CopyCtor on macOS would emit a redundant .clone() in the
  // generated Rust. cpp2rust suppresses it so the output matches Linux.
  const_it == end ? 0 : 1;
```

This only happens on iterator types because the distinction between iterator and const_iterator. Other containers don't have this issue.

### DIFF
--- a/tests/unit/out/unsafe/map.rs
+++ b/tests/unit/out/unsafe/map.rs
@@ -91,7 +91,7 @@ unsafe fn main_0() -> i32 {
         UnsafeMapIterator::find_key(&m as *const BTreeMap<i16, Box<u32>>, &4_i16)
             != UnsafeMapIterator::end(&m as *const BTreeMap<i16, Box<u32>>)
     );
-    UnsafeMapIterator::erase(&(*r) as *const BTreeMap<i16, Box<u32>>, &it4);
+    UnsafeMapIterator::erase(&(*r) as *const BTreeMap<i16, Box<u32>>, &it4.clone());
     assert!((((*r).len() as u64) == (3_u64)));
     assert!(
         UnsafeMapIterator::find_key(&m as *const BTreeMap<i16, Box<u32>>, &4_i16)
--- a/tests/unit/out/unsafe/map.rs
+++ b/tests/unit/out/unsafe/map.rs
@@ -50,10 +50,10 @@ unsafe fn main_0() -> i32 {
     let mut it: UnsafeMapIterator<i16, u32> =
         UnsafeMapIterator::find_key(&m as *const BTreeMap<i16, Box<u32>>, &1_i16);
     let mut const_it: UnsafeMapIterator<i16, u32> =
-        UnsafeMapIterator::find_key(&m as *const BTreeMap<i16, Box<u32>>, &10_i16);
+        UnsafeMapIterator::find_key(&m as *const BTreeMap<i16, Box<u32>>, &10_i16).clone();
     let mut x1: u32 = if it == end { 0_u32 } else { *it.second() };
     assert!(((x1) == (4_u32)));
-    let mut x2: u32 = if const_it == end {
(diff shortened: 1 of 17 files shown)


tests/unit/out/unsafe/map.rs
```diff
@@ -65,11 +65,12 @@ unsafe fn main_0() -> i32 {
         *it.second()
     };
     assert!(((x3) == (4_u32)));
-    let mut x4: u32 = if const_it == UnsafeMapIterator::end(&m as *const BTreeMap<i16, Box<u32>>) {
-        0_u32
-    } else {
-        *const_it.second()
-    };
+    let mut x4: u32 =
+        if const_it == UnsafeMapIterator::end(&m as *const BTreeMap<i16, Box<u32>>).clone() {
```
nunoplopes, line 69: this patch says "suppress clone", but it's regressing and adding a lot of clones.

lucic71: Those are ok to add clones that the translator missed. In this specific example, operator== expects both arguments to be const_iterator. The second argument is converted from iterator to const_iterator using the `a0.clone()` rule.

Even though the above example does not need clone, consider the following case in redudant_copy_in_conversion.cpp:

```cpp
static int sink(std::map<int, int>::iterator it) {
  std::map<int, int>::const_iterator cit(it); // <-- iterator -> const_iterator conversion
  return cit == it ? it->second : 0;          // <-- const_iterator == iterator comparisson
}
```

Without cloning in the iterator -> const_iterator conversion, the code would become:

```rs
  let mut cit: UnsafeMapIterator<i32, i32> = it;
  return if cit == it { *it.second() } else { 0 };
```

And trigger the following error:

```
 9 | pub unsafe fn sink_0(mut it: UnsafeMapIterator<i32, i32>) -> i32 {
   |                      ------ move occurs because `it` has type `MapIter<i32, *const BTreeMap<i32, Box<i32>>>`, which does not implement the `Copy` trait                                  
10 |     let mut cit: UnsafeMapIterator<i32, i32> = it;                                                                                                                                       
   |                                                -- value moved here      
11 |     return if cit == it { *it.second() } else { 0 };                                                                                                                                     
   |                      ^^ value borrowed here after move                                    
   |                                                                                                                                                                                          
help: consider cloning the value if the performance cost is acceptable                         
   |
10 |     let mut cit: UnsafeMapIterator<i32, i32> = it.clone();
   |                                                  ++++++++
```

The solution is to clone the value when performing iterator -> const_iterator conversion:

```rs
  let mut cit: UnsafeMapIterator<i32, i32> = it.clone();
  return if cit == it.clone() { *it.second() } else { 0 };
```

The suppressed clones are in iterator conversion as described in the initial comment: 

> From, tests/unit/redundant_copy_in_conversion:
> 
> ```c++
>   // Comparing const_iterator with iterator forces an implicit conversion of
>   // `end` to const_iterator. The AST shape differs between platforms:
>   //
>   //   Linux: const_it == ConvertingCtor(end)
>   //   macOS: const_it == ConvertingCtor(CopyCtor(end))
>   //
>   // The extra inner CopyCtor on macOS would emit a redundant .clone() in the
>   // generated Rust. cpp2rust suppresses it so the output matches Linux.
>   const_it == end ? 0 : 1;
> ```
> 
> This only happens on iterator types because the distinction between iterator and const_iterator. Other containers don't have this issue.

PR #120, opened 2026-05-18
Rewrite compound assignment in a single expression

lucic71:


nunoplopes:
Uhm, it seems that computing the ptr and storing it in a temporary variable is a good thing for long expressions. This change regresses the performance in those cases. Can we keep that part?

lucic71:
> Uhm, it seems that computing the ptr and storing it in a temporary variable is a good thing for long expressions. This change regresses the performance in those cases. Can we keep that part?

So it should be:

```rs
{
  let _ptr = ptr.clone;
  _ptr.write(_ptr.read() + 1)
}
```

?

lucic71:
> > Uhm, it seems that computing the ptr and storing it in a temporary variable is a good thing for long expressions. This change regresses the performance in those cases. Can we keep that part?
> 
> So it should be:
> 
> ```rust
> {
>   let _ptr = ptr.clone;
>   _ptr.write(_ptr.read() + 1)
> }
> ```
> 
> ?

I did this

cpp2rust/converter/models/converter_refcount.cpp
```diff
@@ -1796,13 +1796,13 @@ void ConverterRefCount::ConvertAssignment(clang::Expr *lhs, clang::Expr *rhs,
...
-      // *ptr += val => { let __ptr = ptr.clone(); let __tmp = __ptr.read() +
-      // val;
-      // __ptr.write(__tmp) }
       auto op = assign_operator;
       op.remove_suffix(1); // remove '='
-      StrCat("{ let __ptr = ", ptr, ".clone(); let __tmp = __ptr.read() ", op,
-             " ", rhs_as_string, "; __ptr.write(__tmp) }");
+      {
+        PushBrace brace(*this);
+        StrCat(std::format("let _ptr = {}.clone();", ptr));
```
nunoplopes, line 1803: why doesn't this use the emitFreshPointer functionality instead of always doing a clone?

lucic71: No good reason. I will replace with ConvertFreshPointer

lucic71: I took a closer look and it's not trivial to replace with ConvertFreshPointer. This path uses the pending_deref_ mechanism. pending_deref_ is generally used for rewriting mutable method receivers to use Ptr::with_mut. Or more generally, to convert pointer dereferences in AddrOf/LValue contexts.

pending_deref_ is triggered inside ConvertLValue, line 1796 above, so every user of ConvertLValue must also check pending_deref_.

PR #121, opened 2026-05-18
Redefine __FILE__ to use only the basename of the file

lucic71:


cpp2rust/converter/converter.cpp
```diff
@@ -275,9 +275,7 @@ bool Converter::VisitUsingType(clang::UsingType *type) {
 bool Converter::Convert(clang::Decl *decl) { return TraverseDecl(decl); }
 
 bool Converter::VisitTranslationUnitDecl(clang::TranslationUnitDecl *decl) {
-  if (auto main_file_name = GetMainFileName(ctx_); main_file_name != "input") {
-    StrCat("\n//", main_file_name + ".rs\n");
-  }
+  StrCat("\n//", GetMainFileName(ctx_) + ".rs\n");
```
nunoplopes: doesn't seem very useful to have this comment. I suggest removing this altogether

lucic71: I dropped the comment by adding a single_file flag

lucic71: Ah sorry, I though you meant reverting to the old behavior. I will delete the comment and the single_file flag

lucic71: Done

PR #123, opened 2026-05-19
Don't map user-defined expressions

lucic71:
If the user code redefines strchr, don't apply the translation rules on the user-defined strchr, instead call the user-defined strchr.

### DIFF
--- a/tests/unit/user_defined_same_as_libc.c
+++ b/tests/unit/user_defined_same_as_libc.c
@@ -1,10 +1,14 @@
 #include <assert.h>
-#include <stddef.h>
+#include <stdio.h>
 
-char *strchr(const char *s, int c) { return NULL; }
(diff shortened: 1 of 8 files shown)


tests/unit/user_defined_same_as_libc.c
```diff
@@ -0,0 +1,14 @@
...
+#include <stdio.h>
+
+FILE *fopen(const char *path, const char *mode) {
+  (void)path;
+  (void)mode;
+  return NULL;
+}
+
+int main() {
+  FILE *fp = fopen("/etc/passwd", "r");
```
nunoplopes: let's avoid this kind of jokes..

lucic71: I changed to /tmp/irrelevant-file

PR #126, opened 2026-05-19
More stdio rules (fputc, fputs, puts, fileno)

lucic71:


tests/unit/stdio.c
```diff
@@ -0,0 +1,44 @@
...
+  puts("puts hello");
+  const char *s = "puts variable";
+  puts(s);
+}
+
+static void test_fileno(void) {
+  assert(fileno(stdin) == 0);
+  assert(fileno(stdout) == 1);
+  assert(fileno(stderr) == 2);
+  FILE *fp = fopen("/tmp/cpp2rust_fileno_test.tmp", "wb");
```
nunoplopes: this is bad; we need to start cleaning trash left behind. Please leave a TODO for whenever unlink works.

lucic71: It will come in #127. I will leave a TODO in this PR

lucic71: Done

PR #130, opened 2026-05-20
[unsafe] Add time.h rules

lucic71:


lucic71:
I will close this for now. On maOS, the rule for gettimeofday does not compile because the second argument is different. On Linux the second argument is a `struct timezone *`, while on macOS this is a `void*`.

```
    --> src/../time/tgt_unsafe.rs:9:28
     |
   9 |     libc::gettimeofday(a0, a1 as *mut ::libc::timezone)
     |     ------------------     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `*mut c_void`, found `*mut timezone`
     |     |
     |     arguments to this function are incorrect
     |
     = note: expected raw pointer `*mut c_void`
                found raw pointer `*mut timezone`
note: function defined here
    --> /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libc-0.2.186/src/unix/bsd/apple/mod.rs:4352:12
     |
4352 |     pub fn gettimeofday(tp: *mut crate::timeval, tz: *mut c_void) -> c_int;
     |            ^^^^^^^^^^^^

```

PR #136, opened 2026-05-21
Take the address of malloc,free,realloc,calloc,strdup

lucic71:


nunoplopes:
Why do we need to move these functions to libcc2rs?

lucic71:
> Why do we need to move these functions to libcc2rs?

To be able to write `void *(*pmalloc)(size_t) = malloc;`. cpp2rust needs to be able to generate

```rs
let mut pmalloc: Option<unsafe fn(u64) -> *mut ::libc::c_void> = Some(libcc2rs::malloc_unsafe);
```

nunoplopes:
Ok, but why can't it write
```rust
let mut pmalloc: Option<unsafe fn(u64) -> *mut ::libc::c_void> = Some(libc::malloc);
```
?

lucic71:
> Ok, but why can't it write
> 
> ```rust
> let mut pmalloc: Option<unsafe fn(u64) -> *mut ::libc::c_void> = Some(libc::malloc);
> ```
> 
> ?

For rules that are translated using a single function call, for example `malloc -> libc::malloc`, it could. I think this requires a new field in the IR to detect this case.

For rules where we write a custom body, for example the previous fwrite: `fwrite -> { custom body that writes *u8 into fs::File }` we need to add a new function in libcc2rs.

Also for refcount translation rules where we won't do `malloc -> libc::malloc` because libc::maloc is unsafe, we will have to add a wrapper function in libcc2rs so that we can take its address.

lucic71:
> For rules that are translated using a single function call, for example `malloc -> libc::malloc`, it could. I think this requires a new field in the IR to detect this case.

Actually, no. I just tested doing this and it errors because the type of maloc is `unsafe extern "C" fn(_) -> _` and the type of the functions that we translate is `unsafe fn(_) -> _`: https://godbolt.org/z/jK8sofzW1

```
  --> <source>:10:73
   |
10 |     let _bad: Option<unsafe fn(usize) -> *mut core::ffi::c_void> = Some(malloc);
   |                                                                    ---- ^^^^^^ expected "Rust" fn, found "C" fn
   |                                                                    |
   |                                                                    arguments to this enum variant are incorrect
   |
   = note: expected fn pointer `unsafe fn(_) -> _`
                 found fn item `unsafe extern "C" fn(_) -> _ {malloc}`
```

I will have to check if it's practical to solve this problem or it's better to keep malloc_unsafe as a libcc2rs function

nunoplopes:
Uhm, ok, then it's an ABI problem. It's a bit annoying if we have to start adding wrappers for every single function in libc. but the alternative would be to have a more heavyweight function pointer that would know how to dispatch calls to different ABI variants.
Let's go with your trampoline option for now.

PR #143, opened 2026-05-22
Add dummy definition for referenced but never-defined structs

lucic71:
This handles opaque pointers to user defined forward declared structs. It's allowed to use such pointers, even though the struct is not defined in any translation unit.

To implement this, I modified record_decls_ to also keep an extra bool with the following meaning:

```
    // record name -> true if a definition has been emitted, false if only
    // referenced.
    std::unordered_map<std::string, bool> entries_;
```

At the end of visiting all TU's, for the record_decls_ that are only referenced (not defined), a dummy struct definition is generated:

```rs
#[repr(C)]
pub struct opaque {
    _opaque: [u8; 0],
}
```

### DIFF
--- a/cpp2rust/converter/converter.cpp
+++ b/cpp2rust/converter/converter.cpp
@@ -23,6 +23,7 @@ std::unordered_set<std::string> Converter::record_decls_;
 std::unordered_set<std::string> Converter::decl_ids_;
 std::unordered_set<std::string> Converter::globals_;
 std::unordered_set<std::string> Converter::abstract_structs_;
+std::unordered_map<std::string, std::string> Converter::referenced_records_;
 
 void Converter::ConvertUniquePtrDeref(clang::CXXOperatorCallExpr *expr) {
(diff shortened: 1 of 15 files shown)


lucic71:
`pub struct opaque` is enough, it doesn't have to be

```rs
#[repr(C)]
pub struct opaque {
    _opaque: [u8; 0],
}
```

nunoplopes:
But we take the whole program as input. Why is something undefined?

lucic71:
> But we take the whole program as input. Why is something undefined?

The following case is a valid C/C++ program:

```c

struct S; // forward declaration

#ifdef FLAG_THAT_CAN_BE_DISABLED_AT_BUILD_TIME
struct S {
  ... definition of struct ...
}
#endif 

void foo(S* s) {}
```

This becomes:

```rs
fn foo(s: *mut S) {}
```

And errors because there is no definition of S (we don't generate code on `struct S; // forward declaration` because rust does not support forward declarations.

PR #152, opened 2026-05-26
Handle non-uniform C++ st_mtim.tv_sec vs Rust st_mtime

lucic71:
`struct stat` contains a field called `st_mtime`. On Linux, this field is defined as a macro expansion to an internal field:

```cpp
struct timespec st_mtim;    /* Time of last modification.  */

#  define st_mtime st_mtim.tv_sec
```

Rust libc does not contain the `st_mtim` field, it only contains `st_mtime`. To solve this tension, I rewrite all encoutners of `::struct stat::st_mtim::tv_sec` as `::libc::stat::st_mtime`.

I didn't want to touch the converter for this change, but I could not find a solution that only involved rules or the compat layer in `cpp2rust/compat`.

### DIFF
--- a/cpp2rust/converter/converter.cpp
+++ b/cpp2rust/converter/converter.cpp
@@ -2556,6 +2556,45 @@ bool Converter::VisitMemberExpr(clang::MemberExpr *expr) {
   return false;
 }
 
+// Returns the inner member and the replacement string.
+static std::pair<clang::MemberExpr *, std::string>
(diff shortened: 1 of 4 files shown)


nunoplopes:
This is ok for now, but ofc it could have been done with rules.

Please add a test.

lucic71:
> This is ok for now, but ofc it could have been done with rules.
> 
> Please add a test.

I added a test

PR #159, opened 2026-05-27
Add translation rules for variadic libc calls

lucic71:
Variadic rules are defined with the following syntax:

```cpp
template <typename... Args>
int f1(int a0, unsigned long a1, Args... args) {
  return ioctl(a0, a1, args...);
}
```

and

```rs
unsafe extern "C" {
    fn f1(a0: i32, a1: u64, ...) -> i32;
}
```

For the C++ side, use a template argument pack. This makes it clear that the function call is variadic. For the Rust side, use a variadic function inside an extern block, that's the only place Rust allows variadic functions.

I added a new attribute in the IR, called is_extern. It's used by the new `Mapper::IsLibcPassthrough`. A libc passthrough is a rule that it's declared in Rust extern block. In the above example ioctl is defined as a passthrough for `libc::ioctl`. In the future, we can use this for non-variadic rules as well.


lucic71:
@joaotgouveia please take a look at the changes in cpp2rust/cpp_rule_preprocessor.cpp

lucic71:

nunoplopes:
Uhm, this seems like a hack that works for a few unsafe rules, but it's of no use for refcount. Doesn't feel like the solution we need.

lucic71:
> Uhm, this seems like a hack that works for a few unsafe rules, but it's of no use for refcount. Doesn't feel like the solution we need.

It works for all variadic unsafe rules. I designed this only with unsafe in mind. Probably for refcount we will need another solution.

Let me think if I can come up with something that would also work for refcount.

cpp2rust/cpp_rule_preprocessor.cpp
```diff
@@ -349,7 +349,10 @@ class Callback : public clang::ast_matchers::MatchFinder::MatchCallback {
   createTemplateArguments(clang::TemplateDecl *decl,
                           llvm::SmallVectorImpl<clang::TemplateArgument> &out) {
     for (clang::NamedDecl *param : *decl->getTemplateParameters()) {
-      if (llvm::isa<clang::TemplateTypeParmDecl>(param)) {
+      if (param->isTemplateParameterPack()) {
+        out.emplace_back(
+            clang::TemplateArgument::CreatePackCopy(sema_->Context, {}));
```
joaotgouveia: I think that this can be replaced by `clang::TemplateArgument::getEmptyPack()`

lucic71: Done

lucic71:
> > Uhm, this seems like a hack that works for a few unsafe rules, but it's of no use for refcount. Doesn't feel like the solution we need.
> 
> It works for all variadic unsafe rules. I designed this only with unsafe in mind. Probably for refcount we will need another solution.
> 
> Let me think if I can come up with something that would also work for refcount.

For unsafe define the rule using a variadic argument syntax:

```rs
unsafe extern "C" {
    fn f21(a0: *mut i8, a1: usize, a2: *const i8, ...) -> i32;
}
```

This means that the translation rule is a direct passthrough to the libc equivalent.

For refcount use a fixed variadic agrument:

```rs
fn f1(a0: i32, a1: u64, va: &[VaArg]) -> i32 {
    panic!(
        "ioctl is not supported in the refcount model (fd={}, request={}, varargs={})",
        a0,
        a1,
        va.len()
    )
}
```

VaArg already is used for translating user-defined variadic functions.

lucic71:
I didn't add proper translation rules for the safe equivalents because that requires more work. For the moment I added a panic implemnetation for each of them. I will add the proper translation later.

PR #163, opened 2026-06-01
Add codegen support for goto

lucic71:
Depends on #161. See #161 for a more detailed description of how goto is implemented.

On the codegen part, this PR emits the goto_block and goto macros when it hits functions that use labels and gotos internally. It also hoists local variables at the start of the function so that all goto_block arms can see and use the local variables.


lucic71:
One other thing worth mentioning: initially I considered not hoisting the variables at the beginning of the functionin the codegen.  Instead I tried to emit the variables as they were declared in the program, and let goto_block do the hoisting inside the macro, however this turned out to be a bit more complicated because I needed to duplicate the GetDefaultAsString logic inside libcc2rs-macro.

nunoplopes:
What happens with VLAs declare inside goto?
e.g.
```
lbl:
 int a[n];
 ...
  ++n;
  goto lbl;
```

lucic71:
> What happens with VLAs declare inside goto? e.g.
> 
> ```
> lbl:
>  int a[n];
>  ...
>   ++n;
>   goto lbl;
> ```

I did not think about that. We don't currently translate VLAs so I skipped this case. Should we add support for that in this PR or do we leave it for later?

PR #164, opened 2026-06-01
Make Rust type rules standalone functions

joaotgouveia:
Changes the format of the Rust side of type rules to:
```rust
fn t<digit>() -> TYPE {
    return DEFAULT-INITIALIZER;
}
```

Actual changes to rules:

- The initializer for the unsafe rules `t1`, `t2`, and `t3` in `fstream` was changed from `::std::fs::File::open("")?` to `::std::fs::File::open("").unwrap()`. This change was made because the `?` operator requires the function to return a value of type `Result` or `Option`. Additionally, other rules for file-related types already use `::std::fs::File::open("").unwrap()` as an initializer (for example, rule `t1` in `iostream`).

- Added a missing generic parameter to the refcount rules `f10` and `f11` in `unique_ptr`. These rules were relying on the struct defined at the top of the type.

### DIFF
--- a/rule-preprocessor/src/syntactic.rs
+++ b/rule-preprocessor/src/syntactic.rs
@@ -537,17 +537,22 @@ impl<'a> TypeIrBuilder<'a> {
             .fn_item
             .ret_type()
             .and_then(|rt| rt.ty())
-            .expect("type rule must declare a return type");
+            .expect("Type rules must declare a return type");
         let (is_refcount_pointer, is_unsafe_pointer) = pointer_flags(&ty);
 
-        let body = self.fn_item.body().expect("type rule must have a body");
-        let init = body
-            .syntax()
-            .descendants()
-            .find_map(ast::ReturnExpr::cast)
-            .and_then(|ret| ret.expr())
-            .or_else(|| body.stmt_list().and_then(|sl| sl.tail_expr()))
-            .expect("type rule must yield an initializer");
+        let stmts = self
+            .fn_item
+            .body()
+            .and_then(|bd| bd.stmt_list())
+            .expect("Types rule must have a body");
+        assert!(
+            stmts.statements().count() == 0,
+            "Type rules mustn't contain anything in the body besides the tail expression"
+        );
+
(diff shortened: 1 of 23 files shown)


rule-preprocessor/src/syntactic.rs
```diff
@@ -531,35 +532,30 @@ impl<'a> TypeIrBuilder<'a> {
...
-                },
-            ));
+        let body = self.fn_item.body().expect("type rule must have a body");
+        let init = body
+            .syntax()
+            .descendants()
+            .find_map(ast::ReturnExpr::cast)
+            .and_then(|ret| ret.expr())
+            .or_else(|| body.stmt_list().and_then(|sl| sl.tail_expr()))
+            .expect("type rule must yield an initializer");
```
lucic71: I think we should add an assert or change init to only accept `{ X }`, not `{ return X; }` or other type of expressions.

lucic71 (commented):
Looks good!

rule-preprocessor/src/syntactic.rs
joaotgouveia: Aren't these two alternatives equivalent in practice? I opted to accept both options because I thought the added flexibility might be useful in the future, particularly when generating these rules automatically. Even if that doesn't turn out to be the case, what would be the disadvantages of allowing both forms?

lucic71: >  Even if that doesn't turn out to be the case, what would be the disadvantages of allowing both forms?

None, `{ X }` just feels more natural

joaotgouveia: Should I change it to only allow that form then?

lucic71: Yes, please. And also make sure that we only support:

```rs
fn t<digit>() -> TYPE {
    DEFAULT-INITIALIZER
}
```

not

```rs
fn t<digit>() -> TYPE {
    expr1;
    expr2;
    ...
    DEFAULT-INITIALIZER
}
```

PR #168, opened 2026-06-03
Wrap variadic arg in paren before calling .into()

lucic71:
`libc::strlen(dummy as *const i8) as u64.into()` (without paren) errors because we can't write `as u64.into()`, we need to wrap the entire arg in a paren before calling .into(): `(libc::strlen(dummy as *const i8) as u64).into()`


nunoplopes:
This one has conflicts

lucic71:
> This one has conflicts

Solved

PR #169, opened 2026-06-03
Handle non-printable ASCII in str and byte arrays

lucic71:
`"\xff"` or `'\xff'` are valid strings/char literals in C/C++. They were not escaped and produced an invalid character that broke the compilation.

After I escaped them, I realized that they cannot be translated to `\xff` in Rust `str`'s because str only accepts `\x00 - \x7f`. Byte array accepts `> \x7f`.

I changed Ptr::from_string_literal to accept a byte array instead of a str. I also changed GetFmtArg to refuse to create Rust `str` that contains chars `> \x7f` and fallback to GetRawArg that produces escaped byte arrays.


lucic71:
@nunoplopes, should we add a github runner that flags usages of single-char string literals?

nunoplopes:
> @nunoplopes, should we add a github runner that flags usages of single-char string literals?

sounds like a bit overkill.. Copilot finds these.

PR #182, opened 2026-06-04
Hoist const variables from goto_block

lucic71:


lucic71:
Hmm, maybe a better strategy is to not translated the hoiste const variable as mut, and instead translate it as const.

nunoplopes:
A variable might be const but still have a non-const initializer, so this is the best we can do for the general case.

PR #183, opened 2026-06-04
Add rules for stdlib, rustls, and stdio

lucic71:
This mainly adds translation rules for rustls_ffi. I needed to do a few modifications in `cpp2rust/` to be able to add translation rules for typedef'ed integers, for example rustls_ffi::rustls_result (`typedef uint32_t rustls_result`):
* add support in cpp-rule-preprocessor to define rule types as `typedef` in C and `using` in C++. rustls.h uses C specific features, hence the source rules must be put in a src.c file. For example in rustls.h uses the following which is valid C but invalid C++:

```c
enum rustls_result { ... }
typedef uint32_t rustls_result; // same name as the enum
```
* zero-initialize rust enums using the enum default value, not using 0, in getIntegerLiteral, for example: `rustls_result r = 0`
* install rustls.h into rules/rustls using the cargo installation instead of copy-pasting the vendored rustls.h file. `rules/build.rs` takes care of this


rules/rustls/rustls.h
```diff
@@ -0,0 +1,2869 @@
+#ifndef RUSTLS_H
+#define RUSTLS_H
```
nunoplopes: why is this file needed and what's the origin?

lucic71: It's needed to get the C symbols that we translate to rust, similar to brotli: https://github.com/Cpp2Rust/cpp2rust/tree/master/rules/brotli/brotli

This is the origin: https://github.com/rustls/rustls-ffi/blob/main/librustls/src/rustls.h

I installed the deb package (https://github.com/rustls/rustls-ffi/releases/tag/v0.15.3) for rustls-ffi on my system and copied /usr/include/rustls.h into rules/rustls/rustls.h

lucic71: I removed the copy-pasted file, see the new description of the PR

CMakeLists.txt
```diff
@@ -171,8 +173,7 @@ file(GLOB_RECURSE rule_preprocessor_sources
      ${PROJECT_SOURCE_DIR}/rule-preprocessor/src/*.rs)
 
 add_custom_command(
-  OUTPUT ${rust_rules_ir_outputs}
-  COMMAND cargo +${RUST_STABLE_VERSION} build --release --manifest-path "${PROJECT_SOURCE_DIR}/rules/Cargo.toml"
```
nunoplopes, line 175: this change doesn't look intended

lucic71: Yes, that was a mistake. It was asymptomatic because I already had rustls.h generated on my machine

PR #187, opened 2026-06-11
Extend cpp-rule-preprocessor supported C++ rules

joaotgouveia:
Extends the set of rules that `cpp-rule-preprocessor` can handle.

The biggest additions are `createMirrorType` and support for adding type hints to IR rules in the form of default template arguments.
One limitation on the kinds of rules that could be processed was that the synthesized PODs did not necessarily exhibit all of the characteristics required to instantiate a given C++ construct.

This patch adds support for specifying default template arguments in rules, which are used as a reference for the characteristics that the synthesized structs must exhibit.
This information is then used by `createMirrorType`.

When the provided type is a non-template class, the synthesized struct simply inherits from it.
When the provided type is a class template, the synthesized struct becomes an instantiation of a synthesized class template that inherits from it.
Roughly:
```cpp
// rule with hint
template <typename T1, typename T2 = std::allocator<T1>>
void f1() { (...) }

// what is roughly synthesized:

// POD, same as before
struct T1 {};

// synthesized class template that inherits from the hint
template <typename _Tp>
class Mirror : public std::allocator<_Tp> {};

// instantiation of the synthesized class template
template <>
class Mirror<T1> {};
```

Using a synthesized class template rather than inheriting directly allows the synthesized struct to match the hinted type structurally.
This is useful when instantiating constructs that perform structural checks, such as:
```cpp
// bits/alloc_traits.h
static_assert(is_same<
  typename __replace_first_arg<_Tp, typename _Tp::value_type>::type, _Tp>::value,
  "allocator_traits<A>::rebind_alloc<A::value_type> must be A");
```

These changes enable `cpp-rule-preprocessor` to parse 479 automatically generated rule sources for `std::vector` and its specializations.

### DIFF
--- a/cpp2rust/converter/mapper.cpp
+++ b/cpp2rust/converter/mapper.cpp
@@ -40,6 +40,7 @@ clang::PrintingPolicy getPrintPolicy() {
   policy.SuppressScope = false;
   policy.FullyQualifiedName = true;
   policy.SuppressUnwrittenScope = true;
+  policy.UsePreferredNames = true;
   return policy;
 }
 
--- a/cpp2rust/cpp_rule_preprocessor.cpp
+++ b/cpp2rust/cpp_rule_preprocessor.cpp
@@ -344,6 +344,91 @@ private:
     }
   }
 
+  clang::QualType
+  getTemplateIdType(clang::ClassTemplateDecl *decl,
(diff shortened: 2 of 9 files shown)


nunoplopes:
This needs to add a few rules that make use of this functionality.

PR #189, opened 2026-06-12
Translate goto label attached to switch case

lucic71:
Translates a label atttached to a switch case, from:

```cpp
  switch (n) {
  target:
  case 0:
    ...
  case 1:
    ...
    goto target;
```

to

```rs
    switch!(match n {
        __v if __v == 0 => 'target: {
            ...
        }
        __v if __v == 1 => {
            ...
            goto!('target);
        }
```

To make this work, I refactored all switch-analysing functions into a single struct, called SwitchArm. Now, VisitSwitchStmt first calls AnalyzeSwitchArms, then iterates over the SwitchArms and generates the corresponding Rust code. Each SwitchArm has a field called label used for generating ` __v if __v == 0 => 'target: {`, i.e. a labeled block in Rust.

The changes in libcc2rs-macros/src/switch.rs are minimal. I added support for accepting a labeled block in the switch! macro.

### DIFF
--- a/cpp2rust/converter/converter_lib.cpp
+++ b/cpp2rust/converter/converter_lib.cpp
@@ -876,20 +876,15 @@ clang::Expr *NormalizeToBool(clang::Expr *expr, clang::ASTContext &ctx) {
       /*BasePath=*/nullptr, clang::VK_PRValue, clang::FPOptionsOverride());
 }
 
-std::vector<clang::SwitchCase *>
-GetTopLevelSwitchCases(clang::SwitchStmt *stmt) {
-  std::vector<clang::SwitchCase *> cases;
-  if (auto *body = llvm::dyn_cast<clang::CompoundStmt>(stmt->getBody())) {
-    for (auto *s : body->body()) {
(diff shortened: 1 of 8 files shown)


cpp2rust/converter/converter_lib.cpp
```diff
@@ -879,16 +848,34 @@ static bool SwitchCaseHasFallthrough(clang::Stmt *stmt) {
...
+    clang::Stmt *inner = s;
+    if (auto *outer = clang::dyn_cast<clang::LabelStmt>(inner)) {
+      label = outer->getDecl()->getName();
+      do {
+        inner = clang::cast<clang::LabelStmt>(inner)->getSubStmt();
+      } while (clang::isa<clang::LabelStmt>(inner));
+    }
+
+    if (auto *sc = clang::dyn_cast<clang::SwitchCase>(inner)) {
+      arms.push_back({{GetLastStmtOfSwitchCase(sc)},
```
nunoplopes: you can use emplace_back without {} (more efficient, constructs in place)

lucic71: Done

PR #192, opened 2026-06-12
Use from_fn for non-Copyable Rust types

lucic71:
In Rust, a struct is Copyable if it explicitly derives or implements Copy. All fields being Copy is not enough.

Add the `derives` attribute in the IR that lists all traits that a mapped Rust type derives. For example rustls_slice_bytes, a struct with Copyable fields, doesn't derive Copy. So `derives` contains only Default.

User defined types and builtin types also have their `derives` field filled. The entire PR is built around the following check:

```cpp
    if (auto *rec = element_type->getAsRecordDecl()) {
      if (!RecordDerivesCopy(rec)) {
        return std::format("std::array::from_fn::<_, {}, _>(|_| {})",
                           size_as_string.c_str(), element_type_as_string);
      }
    }
```

Arrays of non-copyable types cannot be initialized with `[non_copyable_default, N]`, they have to use the `std::array:from_fn` format.

### DIFF
--- a/tests/unit/out/unsafe/array_of_noncopy_struct.rs
+++ b/tests/unit/out/unsafe/array_of_noncopy_struct.rs
@@ -1,0 +1,30 @@
+extern crate libc;
+use libc::*;
+extern crate libcc2rs;
+use libcc2rs::*;
+use std::collections::BTreeMap;
+use std::io::{Read, Seek, Write};
+use std::os::fd::{AsFd, FromRawFd, IntoRawFd};
+use std::rc::Rc;
+#[repr(C)]
+#[derive(Clone, Default)]
+pub struct NonCopy {
+    pub data: Vec<i32>,
+    pub tag: i32,
+}
+pub fn main() {
+    unsafe {
+        std::process::exit(main_0() as i32);
+    }
+}
+unsafe fn main_0() -> i32 {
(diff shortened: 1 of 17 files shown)


tests/unit/out/unsafe/array_of_noncopy_struct.rs
```diff
@@ -0,0 +1,30 @@
...
+    pub data: Vec<i32>,
+    pub tag: i32,
+}
+pub fn main() {
+    unsafe {
+        std::process::exit(main_0() as i32);
+    }
+}
+unsafe fn main_0() -> i32 {
+    let mut arr: [NonCopy; 3] = std::array::from_fn::<_, 3, _>(|_| <NonCopy>::default());
```
nunoplopes, line 21: Uhm, I think we can use just `let _arr: [NonCopy; 3] = Default::default();`. No need for `from_fn`?

lucic71: Unfortunately no, Default::default() only works for array sizes < 32: https://godbolt.org/z/c5T6boGbq

nunoplopes: Ahah, ok

PR #195, opened 2026-06-13
Fix class template translations

joaotgouveia:
Fixes how class templates are translated.

Class templates are currently monomorphized:
https://github.com/Cpp2Rust/cpp2rust/blob/8f79a34db82d97f1b41ac2576b8a202458d029ea/cpp2rust/converter/converter.cpp#L3316-L3321

However, the name of each translated Rust struct was the plain C++ name, leading to collisions.

Additionally, the C++ name used in the translation rule inserted into Mapper was also the plain C++ name.
This name does not match how the type itself is printed, which led to internal cpp2rust assertion failures due to missing translation rules.

### DIFF
--- a/cpp2rust/converter/mapper.cpp
+++ b/cpp2rust/converter/mapper.cpp
@@ -723,9 +723,30 @@ bool ParamIsPointer(const clang::Expr *expr, unsigned index) {
   return GetParamInfo(expr, index).is_pointer();
 }
 
+clang::QualType GetTypeForDecl(const clang::NamedDecl *decl) {
+  if (const auto *spec =
+          llvm::dyn_cast<clang::ClassTemplateSpecializationDecl>(decl)) {
+    llvm::ArrayRef<clang::TemplateArgument> args =
+        spec->getTemplateArgs().asArray();
+    llvm::SmallVector<clang::TemplateArgument, 4> canon(args.begin(),
+                                                        args.end());
+    ctx_->canonicalizeTemplateArguments(canon);
+
+    return ctx_->getTemplateSpecializationType(
+        clang::ElaboratedTypeKeyword::None,
+        clang::TemplateName(spec->getSpecializedTemplate()), args, canon);
+  }
+
+  const auto *rdecl = llvm::dyn_cast<clang::TagDecl>(decl);
+  assert(rdecl && "Unsupported decl type");
+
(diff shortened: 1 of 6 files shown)


tests/unit/class_templates.cpp
```diff
@@ -0,0 +1,43 @@
...
+  MyContainer<char> cmc;
+  assert(cmc.empty());
+  cmc.push_back('a');
+  assert(cmc.size() == 1 && cmc.back() == 'a');
+  cmc.pop_back();
+  assert(cmc.empty());
+
+  MyContainer<float> fmc;
+  assert(fmc.empty());
+  fmc.push_back('a');
```
lucic71: 1.0?

cpp2rust/converter/mapper.cpp
```diff
@@ -783,6 +788,16 @@ void AddRuleForUserDefinedType(clang::NamedDecl *decl) {
   }
 }
 
+std::string ToRustName(std::string name) {
+  size_t pos = 0;
+  std::string chars = "<>, ";
```
lucic71: Maybe Mapper::ToString is a better place to do these replaces. We already have some similar replaces that we do to normalize the output

joaotgouveia: `ToString` handles converting a given C++ construct to its corresponding string representation, whereas `ToRustName` is responsible for transforming the string representation of an arbitrary C++ construct into a valid Rust name. Given this distinction, wouldn't it be clearer to keep these two functions separate?

Additionally, this normalization is only required when we want to use the string representation of a given C++ construct as a Rust name. It is not required for the other uses of `ToString`, such as obtaining the string representation of a construct in order to look up or insert a translation rule.

cpp2rust/converter/mapper.cpp
```diff
@@ -783,6 +788,16 @@ void AddRuleForUserDefinedType(clang::NamedDecl *decl) {
...
+  while ((pos = name.find_first_of(chars, pos)) != std::string::npos) {
+    name.replace(pos, 1, 1, '_');
```
nunoplopes: name[pos] = '_';

cpp2rust/converter/mapper.cpp
nunoplopes: remove this var to avoid allocations (inline in find_first_of)

PR #203, opened 2026-06-23
Avoid hoisting for function call arguments that don't alias

lucic71:
In terms of converter, the PR is small. I moved GetAllVars and MayCauseBorrowMutError from converter to converter_lib because they are converter-agnostic. Then I added the aliasing logic in CollectCallInfo 2 functions in converter_lib: ReferencesThis and ArgsMayAlias.

The core idea is: if an argument of a function call does not alias with any other argument, then it can be inlined instead of hoisted. This includes aliases over this, for example `this.method(this.member)`

### DIFF
--- a/cpp2rust/converter/converter_lib.cpp
+++ b/cpp2rust/converter/converter_lib.cpp
@@ -644,25 +644,28 @@ clang::Expr *GetCallObject(clang::CallExpr *expr) {
   return nullptr;
 }
 
-std::unordered_set<const clang::ValueDecl *>
-GetAllVars(const clang::Stmt *stmt) {
-  std::unordered_set<const clang::ValueDecl *> vars;
+static void GetAllVarsImpl(const clang::Stmt *stmt,
+                           std::unordered_set<const clang::ValueDecl *> &vars) {
   if (!stmt) {
-    return vars;
+    return;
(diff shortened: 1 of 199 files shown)


cpp2rust/converter/converter_lib.cpp
```diff
@@ -643,6 +644,75 @@ clang::Expr *GetCallObject(clang::CallExpr *expr) {
...
+  std::unordered_set<const clang::ValueDecl *> vars;
+  if (!stmt) {
+    return vars;
+  }
+
+  if (auto *decl_ref = clang::dyn_cast<clang::DeclRefExpr>(stmt)) {
+    vars.insert(decl_ref->getDecl());
+  } else if (auto *member = clang::dyn_cast<clang::MemberExpr>(stmt)) {
+    vars.insert(member->getMemberDecl());
+    auto child_vars = GetAllVars(member->getBase());
```
nunoplopes: It's better to split this method in two: one that returns the set (just calls the 2nd function), and another that returns void and received an additional set parameter and adds stuff there.
This avoid allocating an extra set on the recursive call.

lucic71: Done

PR #205, opened 2026-06-25
Add safe translation for unions

lucic71:
C/C++ unions are translated as a byte buffer + accessors that return a reinterpreted pointer into the byte buffer.

This C/C++ union:

```cpp
union basic {
  int i;
  float f;
};
```

becomes:

```rs
pub struct basic {
    __bytes: Value<Box<[u8]>>,
}
impl basic {
    pub fn i(&self) -> Ptr<i32> {
        (self.__bytes.as_pointer() as Ptr<u8>).reinterpret_cast()
    }
    pub fn f(&self) -> Ptr<f32> {
        (self.__bytes.as_pointer() as Ptr<u8>).reinterpret_cast()
    }
}
```

Where __bytes is the byte buffer and i() and f() are the accessors that return reinterpreted views into the backing buffer.

Compared to Rust unsafe/C/C++ unions, this implementation is safe because it always serializes/deserializes the bytes into valid values or panics otherwise.


cpp2rust/converter/models/converter_refcount.cpp
```diff
@@ -469,6 +482,40 @@ void ConverterRefCount::AddDefaultTrait(const clang::RecordDecl *decl) {
 }
 
 void ConverterRefCount::AddDefaultTraitForUnion(const clang::RecordDecl *decl) {
+  auto name = GetRecordName(decl);
+  StrCat(std::format("impl Default for {}", name));
+  PushBrace impl_brace(*this);
+  StrCat("fn default() -> Self");
+  PushBrace fn_brace(*this);
+  StrCat(std::format("{} {{ __bytes: Rc::new(RefCell::new(vec![0u8; "
```
nunoplopes: I think `Box::from([0u8; {}]) ` is more efficient

cpp2rust/converter/models/converter_refcount.cpp
```diff
@@ -441,17 +441,30 @@ void ConverterRefCount::ConvertOrdAndPartialOrdTraits(
                                     second_return, GetRecordName(decl));
 }
 
-void ConverterRefCount::AddCloneTrait(const clang::CXXRecordDecl *decl) {
-  if (decl->defaultedCopyConstructorIsDeleted()) {
+void ConverterRefCount::AddCloneTrait(const clang::RecordDecl *decl) {
+  auto record_name = GetRecordName(decl);
+
+  if (decl->isUnion()) {
+    StrCat(std::format("impl Clone for {}", record_name));
```
nunoplopes: a lot of these formats are wasteful. It's better to do `StrCat("impl Clone for", record_name)`

cpp2rust/converter/models/converter_refcount.cpp
lucic71: They should be the same: https://godbolt.org/z/z59EYY6Wb. Historically `Box::from([0u8; {}])` made a stack allocation for the array then copied the array on heap. But that seems to be solved so I will use your suggestion

lucic71: Done

cpp2rust/converter/models/converter_refcount.cpp
lucic71: Done

PR #207, opened 2026-06-25
Memcpy a pointer to a struct through ErasedPtr

lucic71:
`AnyPtr::reinterpret_cast` was matching against a hardcoded list of types. This PR adds ErasedPtr::as_bytes which is a `reinterpret_cast::<u8>` so that memcpy/memset a pointer to a struct works.

I also simplified the implementation for ErasedPtr::equals in this refactor.


lucic71:

PR #210, opened 2026-06-29
Handle explicit cast of union member

lucic71:
Union members are translated using accessors:

```cpp
union u { int a; }
*(char *)&u.a = 0xFF;
```
becomes

```rs
struct u { __bytes: Value<Box<[u8]>> }
impl u { pub fn a() -> Ptr<i32> { ... } } // the accessor

u.a().reinterpret_cast<u8>().write(0xFF) // Before this PR it did not compile: (u.a() as Ptr<u8>).write(0xFF)
```

This PR replaces the wrong `(u.a() as Ptr<u8>)` pattern with `u.a().reinterpret_cast<u8>()`


lucic71:
This has no effect on the existing union tests now, but it's a step forward to unblock them.

PR #214, opened 2026-06-30
Add panic-ub test mode

lucic71:
The `panic` mode is currently used for both describing UB tests (e.g. translate null ptr deref as panic) and for describing translation gaps (panic in libcc2rs for types that don't implement ByteRepr).

I renamed the first type as `panic-ub` and the second type as `panic`. So that the `Expectedly Failed Tests` in lit includes the second type, i.e. translation gaps.


nunoplopes:
Can't you use XFAIL for these?

lucic71:
> Can't you use XFAIL for these?

XFAIL is too general; using panic, no-compile, translation-fail it's easier to track the state of a unit test.

PR #218, opened 2026-07-01
Translate access of union array members as `Ptr<T>` instead of `Ptr<Box<[T]>>`

lucic71:

PR #221, opened 2026-07-01
Use reinterpret_sized for array translation in unions

lucic71:
`Box<[T]>::byte_size` cannot be implemented because `[T]` is not `Sized`. For dynamic types, like `Box<[T]>` use `Ptr::reinterpret_sized(usize)` instead of `Ptr::reinterpret_cast()`. The usize argument of reinterpret_sized encodes the C size of the dynamic Rust object.

Internally, `reinterpret_cast` calls `reinterpret_sized(U::byte_size())`.

`Box<[T]>::to_bytes` and `Box<[T]>::from_bytes` can be implemented and they are required in order to read and write a reinterpreted array.

### DIFF
--- a/cpp2rust/converter/converter_lib.cpp
+++ b/cpp2rust/converter/converter_lib.cpp
@@ -150,6 +150,16 @@ bool IsCharArrayFieldFromLibc(const clang::ValueDecl *decl) {
       field->getParent()->getLocation());
 }
 
+bool IsUnionArrayMember(const clang::Expr *base) {
+  if (auto *me =
+          clang::dyn_cast<clang::MemberExpr>(base->IgnoreParenImpCasts())) {
+    if (auto *fd = clang::dyn_cast<clang::FieldDecl>(me->getMemberDecl())) {
+      return fd->getParent()->isUnion() && fd->getType()->isArrayType();
+    }
+  }
+  return false;
+}
+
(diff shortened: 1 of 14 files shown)


nunoplopes:
I'm not following this set of patches.
Shouldn't `raw_()`return `Ptr<u8>` instead of `Ptr<Box<[u8]>>`?

The pointer can access the byte size. Rust knows the size, so I don't understand why we need to store it ourselves.

lucic71:
> I'm not following this set of patches.

After #205 in which I translated unions as a byte array + Ptr accessors, I focused on adding ByteRepr for types that can take part in unions: enums, structs, pointers, arrays.

The current PR adds translation for array members. But I think I over-engineered this. `Ptr<T>` instead of `Ptr<Box<[u8]>>` is enough. My reasoning was that the accessor must return a pointer to the declared member type. But in the case of arrays, returning `Ptr<T>` is easier and actually faster because `Ptr<T>` doesn't have to materialize the entire array, just one element.

> Shouldn't `raw_()`return `Ptr<u8>` instead of `Ptr<Box<[u8]>>`?
> 
> The pointer can access the byte size. Rust knows the size, so I don't understand why we need to store it ourselves.

I will close this PR and add one more small commit over #218 to have array members translated as `Ptr<T>`

PR #226, opened 2026-07-02
Allow passing a specific directory to rule-preprocessor

joaotgouveia:
Adds an optional command-line argument that allows specifying a directory other than the default `../rules`.

### DIFF
--- a/rule-preprocessor/src/main.rs
+++ b/rule-preprocessor/src/main.rs
@@ -19,11 +19,13 @@ use semantic::SemanticAnalysis;
 use syntactic::SyntacticAnalysis;
 
 fn main() {
-    let out_dir = std::env::args()
-        .nth(1)
-        .expect("usage: rule-preprocessor <out-dir>");
+    let mut args = std::env::args().skip(1);
+    let out_dir = args
+        .next()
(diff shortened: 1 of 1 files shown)


rule-preprocessor/src/main.rs
```diff
@@ -19,11 +19,13 @@ use semantic::SemanticAnalysis;
...
 
 fn main() {
-    let out_dir = std::env::args()
-        .nth(1)
-        .expect("usage: rule-preprocessor <out-dir>");
+    let mut args = std::env::args().skip(1);
+    let out_dir = args
+        .next()
+        .expect("usage: rule-preprocessor <out-dir> [rules-crate-dir]");
+    let in_dir = args.next().unwrap_or_else(|| "../rules".to_string());
```
lucic71, line 26: I think you should terminate the program if the dir is not found. Otherwise it will create confusion when someone wants to use dir X and instead it fallbacks to rules/

joaotgouveia: The fallback to `../rules` only happens when the argument is not provided, i.e., it acts as a default argument. If the provided directory is not found, the program terminates with a `No such file or directory` error.

PR #229, opened 2026-07-06
Add Ptr::from_int and Ptr::to_int

lucic71:


lucic71:
Hmm, now that I think about it more, to_int can return usize and from_int can receive usize

lucic71:
> Hmm, now that I think about it more, to_int can return usize and from_int can receive usize

Done

PR #231, opened 2026-07-06
Add void pointer implicit conversion

lucic71:


tests/unit/out/refcount/void_ptr_implicit_conversion.rs
```diff
@@ -0,0 +1,34 @@
...
+use std::cell::RefCell;
+use std::collections::BTreeMap;
+use std::io::prelude::*;
+use std::io::{Read, Seek, Write};
+use std::os::fd::AsFd;
+use std::rc::{Rc, Weak};
+pub fn bump_0(arg: AnyPtr) -> i32 {
+    let arg: Value<AnyPtr> = Rc::new(RefCell::new(arg));
+    let value: Value<Ptr<i32>> = Rc::new(RefCell::new(
+        ((*arg.borrow()).reinterpret_cast::<i32>()).clone(),
```
nunoplopes, line 12: why is this clone appearing here?

lucic71: It comes from ConvertVarInit. Initializers of variables are cloned in order to avoid moves

PR #234, opened 2026-07-06
Add safe malloc/calloc/realloc/free/strdup

lucic71:


lucic71:
I will add the proper implementation of those functions in this PR and I will change the name of this PR

PR #250, opened 2026-07-13
Add safe cstring rules

lucic71:


lucic71:
@nunoplopes do you think it's better to implement most of the cstring functions as part of `Ptr<u8>`? The current rules are not very idiomatic

rules/cstring/tgt_refcount.rs
```diff
@@ -18,14 +18,414 @@ fn f3(a0: AnyPtr, a1: AnyPtr, a2: usize) -> i32 {
 }
 
 fn f4(a0: AnyPtr, a1: AnyPtr, a2: usize) -> AnyPtr {
-    a0.memcpy(&a1, a2 as usize);
+    let __tmp: Vec<u8> = (0..a2)
```
nunoplopes: why can't we use memcpy here?

rules/cstring/tgt_refcount.rs
```diff
@@ -18,14 +18,414 @@ fn f3(a0: AnyPtr, a1: AnyPtr, a2: usize) -> i32 {
...
+    for __i in 0..a2 {
+        a0.reinterpret_cast::<u8>().offset(__i).write(__tmp[__i]);
+    }
     a0.clone()
 }
 
+fn f5(a0: Ptr<u8>, a1: i32) -> Ptr<u8> {
+    let mut __i: usize = 0;
+    loop {
+        let __c = a0.offset(__i).read();
```
nunoplopes: this is inefficient as it's creating a new weak ref on each iteration. Do a ptr clone before the loop, and increment the pointer with += 1.

rules/cstring/tgt_refcount.rs
```diff
@@ -18,14 +18,414 @@ fn f3(a0: AnyPtr, a1: AnyPtr, a2: usize) -> i32 {
...
+            break Ptr::null();
+        }
+        __i += 1;
+    }
+}
+
 unsafe fn f7(a0: Ptr<u8>) -> usize {
-    a0.to_string_iterator().count()
+    let mut __i: usize = 0;
+    while a0.offset(__i).read() != 0 {
```
nunoplopes: same comment here. Can use pointer arithmetic and subtract the final and initial pointers

rules/cstring/tgt_refcount.rs
```diff
@@ -18,14 +18,414 @@ fn f3(a0: AnyPtr, a1: AnyPtr, a2: usize) -> i32 {
...
+        __i += 1;
+    }
+    __i
+}
+
+fn f8(a0: Ptr<u8>, a1: Ptr<u8>) -> i32 {
+    let mut __i: usize = 0;
+    loop {
+        let __c1 = a0.offset(__i).read();
```
nunoplopes: same comment; use pointer arithmetic

rules/cstring/tgt_refcount.rs
lucic71: f4 is memmove. memmove allows overlap between src and dst, memcpy does not (it's UB). Maybe it would be a good idea to add an assert inside Ptr::memcpy to panic on overlapping regions.

rules/cstring/tgt_refcount.rs
lucic71: Done

rules/cstring/tgt_refcount.rs
lucic71: Done

rules/cstring/tgt_refcount.rs
lucic71: Done

rules/cstring/tgt_refcount.rs
nunoplopes: But our implementation of memcpy is ok with overlaps. There's no point in reinventing the wheel.

rules/cstring/tgt_refcount.rs
```diff
@@ -18,14 +18,474 @@ fn f3(a0: AnyPtr, a1: AnyPtr, a2: usize) -> i32 {
...
+            break __p;
+        }
+        if __c == 0 {
+            break Ptr::null();
+        }
+        __p += 1;
+    }
+}
+
 unsafe fn f7(a0: Ptr<u8>) -> usize {
```
nunoplopes: unsafe

rules/cstring/tgt_refcount.rs
```diff
@@ -18,14 +18,474 @@ fn f3(a0: AnyPtr, a1: AnyPtr, a2: usize) -> i32 {
...
+        __p2 += 1;
+    }
+}
+
+fn f9(a0: Ptr<u8>, a1: Ptr<u8>, a2: usize) -> i32 {
+    let mut __p1 = a0.clone();
+    let mut __p2 = a1.clone();
+    let mut __i: usize = 0;
+    loop {
+        if __i == a2 {
```
nunoplopes: you can decrement a2; there's no need for __i

rules/cstring/tgt_refcount.rs
lucic71: It's not ok if dst > src. For the following code:

```cpp
#include <string.h>

int main(void) {
    char buf[6] = {1, 2, 3, 4, 5, 6};

    memmove(buf + 2, buf, 4);

    for (int i = 0; i < 6; i++) {
      printf("%d ", buf[i]);
    }
    printf("\n");

    return 0;
}
```

The output is:

```
C:                                 1 2 1 2 3 4 (correct, it's a shift right)
memmove translated as Ptr::memcpy: 1 2 1 2 1 2 (wrong)
```

rules/cstring/tgt_refcount.rs
```diff
@@ -18,14 +18,474 @@ fn f3(a0: AnyPtr, a1: AnyPtr, a2: usize) -> i32 {
...
+        __p2 += 1;
+        __i += 1;
+    }
+}
+
+fn f10(a0: AnyPtr, a1: i32, a2: usize) -> AnyPtr {
+    let mut __p = a0.reinterpret_cast::<u8>();
+    let mut __i: usize = 0;
+    loop {
+        if __i == a2 {
```
nunoplopes, line 79: same here; can be decremented

rules/cstring/tgt_refcount.rs
lucic71: I think an assert on overlapping Ptr::memcpy would be useful. Memcpy with overlapping memory is UB in C

rules/cstring/tgt_refcount.rs
```diff
@@ -18,14 +18,474 @@ fn f3(a0: AnyPtr, a1: AnyPtr, a2: usize) -> i32 {
...
+fn f16(a0: Ptr<u8>, a1: Ptr<u8>) -> usize {
+    let mut __p = a0.clone();
+    let mut __i: usize = 0;
+    loop {
+        let __c = __p.read();
+        if __c == 0 {
+            break __i;
+        }
+        let mut __q = a1.clone();
+        let __hit = loop {
```
nunoplopes: I would rather read a1 into a local vector. seems easier to optimize.

rules/cstring/tgt_refcount.rs
```diff
@@ -18,14 +18,474 @@ fn f3(a0: AnyPtr, a1: AnyPtr, a2: usize) -> i32 {
...
+            break __i;
+        }
+        __p += 1;
+        __i += 1;
+    }
+}
+
+fn f18(a0: Ptr<u8>, a1: Ptr<u8>) -> Ptr<u8> {
+    let mut __p = a0.clone();
+    loop {
```
nunoplopes, line 131: this code should be using CStringIterator instead. becomes a lot smaller.

rules/cstring/tgt_refcount.rs
nunoplopes: I know it's UB in C. But our safe Rust implementation can handle overlapping memory. So let's use it.

lucic71: In this case I will change the implementation of Ptr::memcpy to fully behave as memmove

rules/cstring/tgt_refcount.rs
lucic71: It makes no difference, a2 is textually replaced inside the rule. We cannot decrement a2 until it reaches 0, we need to save it in a variable eitherway

rules/cstring/tgt_refcount.rs
lucic71: It makes no difference, a2 is textually replaced inside the rule. We cannot decrement a2 until it reaches 0, we need to save it in a variable eitherway

rules/cstring/tgt_refcount.rs
lucic71: Done

rules/cstring/tgt_refcount.rs
lucic71: Done. Also used CStringIterator in other rules

rules/cstring/tgt_refcount.rs
lucic71: No longer relevant. I replaced the __hit = loop idiom with CStringIterator

rules/cstring/tgt_refcount.rs
lucic71: Done

libcc2rs/src/rc.rs
```diff
@@ -952,6 +952,16 @@ thread_local! {
 impl Ptr<u8> {
     #[allow(clippy::explicit_counter_loop)]
     pub fn memcpy(&self, src: &Self, len: usize) {
+        if *self > *src {
```
nunoplopes, line 955: uhm, can't we change the loop below to increment src and just iterate over o..len?

PR #259, opened 2026-07-16
Add unistd safe rules

lucic71:


libc-dep/Cargo.toml
```diff
@@ -5,3 +5,4 @@ edition = "2024"
 
 [dependencies]
 libc = "0.2"
+nix = { version = "0.30.1", features = ["socket", "net", "fs", "poll", "time", "user", "dir", "term", "event", "hostname"] }
```
nunoplopes: is this dependency needed here?

rules/Cargo.toml
```diff
@@ -13,3 +13,4 @@ libc = "0.2.148"
 libcc2rs = { version = "0.1.0", path = "../libcc2rs" }
 brotli-sys = "0.3"
 rustls-ffi = { version = "0.15.3", default-features = false }
+nix = { version = "0.30.1", features = ["socket", "net", "fs", "poll", "time", "user", "dir", "term", "event", "hostname"] }
```
nunoplopes: drop the .1

libcc2rs/Cargo.toml
```diff
@@ -6,3 +6,4 @@ edition = "2024"
 [dependencies]
 libcc2rs-macros = { path = "../libcc2rs-macros", version = "0.1.0" }
 libc = "0.2"
+nix = { version = "0.30.1", features = ["socket", "net", "fs", "poll", "time", "user", "dir", "term", "event", "hostname"] }
```
nunoplopes: I meant this one is not needed

lucic71: Both libcc2rs and libc-dep depend on libc. Adding nix to libc-dep pulls libc with extra traits. This creates a mismatch between the libc version that libcc2rs uses and the one that libc-dep uses.

To unify the libc versions from libcc2rs and libc-dep, I added nix as a dependency in libcc2rs as well.

Without this, weird errors like the following one appear:

```
   --> main.rs:40:30                       
    |                                                                                          
 40 |     assert!(((((libc::fileno(libcc2rs::stderr_unsafe())) == (2)) as i32) != 0));
    |                 ------------ ^^^^^^^^^^^^^^^^^^^^^^^^^ expected `FILE`, found `libc::unix::FILE`
    |                 |
    |                 arguments to this function are incorrect
    |
note: there are multiple different versions of crate `libc` in the dependency graph
   --> /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libc-0.2.186/src/macros.rs:259:9
    |
259 |           pub enum $i { }
    |           --------------- this is the found type
    |
   ::: /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libc-0.2.186/src/macros.rs:259:9
    |
259 |           pub enum $i { }
    |           ^^^^^^^^^^^^^^^ this is the expected type
    |
   ::: /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libc-0.2.186/src/unix/mod.rs:596:1
    |
596 | / extern_ty! {
597 | |     pub enum FILE {}
598 | | }
    | |_- in this macro invocation
    = help: you can use `cargo tree` to explore your dependency tree
```

rules/Cargo.toml
lucic71: Done

PR #260, opened 2026-07-16
Add safe rules for time

lucic71:
This also adds a new dependency called jiff. jiff is used to avoid manually implementing strftime and localtime_r

In general, nix only contains safe wrappers over syscalls. Everything which is a pure library implementation is not included


rules/time/tgt_refcount.rs
```diff
@@ -14,3 +14,151 @@ fn t2() -> libcc2rs::Timeval {
...
+    match nix::sys::stat::utimes(a0.to_rust_string().as_str(), &__at, &__mt) {
+        Ok(()) => 0,
+        Err(__e) => {
+            libcc2rs::cpp2rust_errno().write(__e as i32);
+            -1
+        }
+    }
+}
+
+#[cfg(target_os = "linux")]
```
nunoplopes, line 113: the implementation seems to be the same in linux and mac; why do we need both?

rules/time/tgt_refcount.rs
```diff
@@ -14,3 +14,151 @@ fn t2() -> libcc2rs::Timeval {
...
+fn f8(a0: Ptr<Timeval>, a1: Ptr<::libc::timezone>) -> i32 {
```
nunoplopes, line 114: with this 2nd parameter?

rules/time/tgt_refcount.rs
lucic71: Only the signature is different between linux and macos. Linux receives a struct timezone pointer, macos receives a void pointer:

https://man7.org/linux/man-pages/man2/gettimeofday.2.html
https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/gettimeofday.2.html

rules/time/tgt_refcount.rs
```diff
@@ -14,3 +14,151 @@ fn t2() -> libcc2rs::Timeval {
...
+        }
+        Err(__e) => {
+            libcc2rs::cpp2rust_errno().write(__e as i32);
+            -1
+        }
+    }
+}
+
+#[cfg(target_os = "macos")]
+fn f8(a0: Ptr<Timeval>, a1: AnyPtr) -> i32 {
```
nunoplopes, line 134: this is ignoring the 2nd parameter. At least it should have a TODO.

lucic71: Both man pages (linux and macos) say that the second param is obsolete. I will add a comment about that

rules/time/tgt_refcount.rs
```diff
@@ -14,3 +14,151 @@ fn t2() -> libcc2rs::Timeval {
...
+                *__tm.tm_min.borrow_mut() = __dt.minute() as i32;
+                *__tm.tm_hour.borrow_mut() = __dt.hour() as i32;
+                *__tm.tm_mday.borrow_mut() = __dt.day() as i32;
+                *__tm.tm_mon.borrow_mut() = __dt.month() as i32 - 1;
+                *__tm.tm_year.borrow_mut() = __dt.year() as i32 - 1900;
+                *__tm.tm_wday.borrow_mut() = __dt.weekday().to_sunday_zero_offset() as i32;
+                *__tm.tm_yday.borrow_mut() = __dt.day_of_year() as i32 - 1;
+                *__tm.tm_isdst.borrow_mut() = 0;
+                *__tm.tm_gmtoff.borrow_mut() = 0;
+                *__tm.tm_zone.borrow_mut() = Ptr::from_string_literal(b"GMT");
```
nunoplopes: should probably be UTC

lucic71: It's GMT on linux and UTC on macos

rules/time/tgt_refcount.rs
lucic71: I added a comment about that

PR #264, opened 2026-07-16
Add misc safe rules

lucic71:


rules/pwd/tgt_refcount.rs
```diff
@@ -6,3 +6,10 @@ use libcc2rs::*;
 fn t1() -> libcc2rs::Passwd {
     Default::default()
 }
+
+fn f1(a0: u32) -> Ptr<Passwd> {
+    match nix::unistd::User::from_uid(nix::unistd::Uid::from_raw(a0)) {
+        Ok(Some(__u)) => Ptr::alloc(Passwd::from_user(&__u)),
+        _ => Ptr::null(),
```
nunoplopes: needs to set errno

rules/locale/tgt_refcount.rs
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2022-present INESC-ID.
+// Distributed under the MIT license that can be found in the LICENSE file.
+
+use libcc2rs::*;
+
+fn f1(a0: i32, a1: Ptr<u8>) -> Ptr<u8> {
+    Ptr::from_string_literal(b"C")
```
nunoplopes, line 11: this is not correct for sure :)

lucic71: Rust does not seem to have [support for localization](https://doc.rust-lang.org/std/fmt/index.html#localization)

The [man page](https://man7.org/linux/man-pages/man3/setlocale.3.html) of setlocale says:

```
       A successful call to setlocale() returns an opaque string that
       corresponds to the locale set.  This string may be allocated in
       static storage.  The string returned is such that a subsequent
       call with that string and its associated category will restore
       that part of the process's locale.  The return value is NULL if
       the request cannot be honored.
```

So the real options here are to return C (`The locale "C" or "POSIX" is a portable locale; it exists on all conforming systems.`) or NULL. NULL looks more like a failure, so I chose "C" instead.

rules/pwd/tgt_refcount.rs
lucic71: Done

rules/locale/tgt_refcount.rs
nunoplopes: We probably need to track the locale ourselves and change the behavior of relevant string functions then

lucic71: I added a TODO about that

PR #287, opened 2026-07-23
Pin ruff version and fix sporadic double unlink in tests

lucic71:
ruff got suddenly updated and our python linting broke. Pin ruff to a fixed version so linting is stable.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -26,7 +26,7 @@ jobs:
           llvm-version: 22
 
       - name: Setup Python
-        run: pip install ruff
+        run: pip install ruff==0.15.22
 
       - name: Configure
         run: cmake -GNinja -B build -S .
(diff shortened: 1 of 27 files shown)


lucic71:
While fixing the ruff version I encountered another problem: C++/unsafe/refcount all work with the same file declared in tmp, for example `const char *path = "/tmp/cpp2rust_lseek_test.tmp";`. This can lead to double unlinks which make tests fail sporadically.

To fix this, I moved all files from /tmp into the temporary build directory of each test, the second argument is cwd:

```diff
-        self.cpp_result = RunResult(*lit.util.executeCommand(str(self.cpp_bin)))
+        self.cpp_result = RunResult(
+           *lit.util.executeCommand(str(self.cpp_bin), str(self.tmp_dir))
+        )
```

PR #288, opened 2026-07-24
Add a `rust-toolchain.toml` for the rule-preprocessor

FranciscoTGouveia:
As this crate requires the `rustc-dev` component, which is not bundled by default, we need to add it explicitly.
Now, a user can just do `cargo build` on this crate.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,7 @@ add_custom_command(
     OUTPUT "${RUST_STAMP_FILE}"
     COMMAND rustup toolchain install ${RUST_STABLE_VERSION} --component clippy --component rustfmt
     COMMAND rustup toolchain install ${RUST_NIGHTLY_VERSION} --component rustc-dev --component llvm-tools --component clippy --component rustfmt
+    COMMAND rustup override set ${RUST_NIGHTLY_VERSION} --path ${PROJECT_SOURCE_DIR}/rule-preprocessor
     COMMAND ${CMAKE_COMMAND} -E touch "${RUST_STAMP_FILE}"
     DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/cmake/rust-toolchain.cmake"
     USES_TERMINAL


FranciscoTGouveia:
In my machine, `ninja format` works just fine; although it seems weird for this to be a spurious failure...
Any ideas @lucic71?

lucic71:
> In my machine, `ninja format` works just fine; although it seems weird for this to be a spurious failure...
> 
> Any ideas @lucic71?

See #287, it's fixed there

FranciscoTGouveia:
> > In my machine, `ninja format` works just fine; although it seems weird for this to be a spurious failure...
> > Any ideas @lucic71?
> 
> See #287, it's fixed there

Great, thanks!
I'll wait for #\287 to be merged then.

rule-preprocessor/rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "nightly-2026-06-30"
```
lucic71: How do you plan to keep this in sync with https://github.com/Cpp2Rust/cpp2rust/blob/master/cmake/rust-toolchain.cmake#L2 ?

nunoplopes:
What's the benefit of this PR?
It's yet another place to go out of sync.

FranciscoTGouveia:
> What's the benefit of this PR?

Currently, if we try to compile the rule-preprocessor as a standalone (i.e., without `ninja build`), it fails, as it does not know which toolchain to use; it needs nightly + the `rustc-dev` component.

> It's yet another place to go out of sync.

I decided to solve this and directly implement the patch on `CMakeLists.txt`.
This way we never get out of sync, and we can compile the crate as a standalone, assuming we've run `ninja build` at least once first.

rule-preprocessor/rust-toolchain.toml
FranciscoTGouveia: Addressed in the comment below.

PR #302, opened 2026-08-14
Add documentation for rules

lucic71:
It covers:
* how to write a rule
* how cpp-rule-preprocessor and rule-preprocessor work
* the format of Rules IR
* etc

See docs/src/SUMMARY.md for a complete list.


lucic71:
@joaotgouveia please also have a look, especially at the cpp-rule-preprocessor part

PR #307, opened 2026-08-16
Fix operator name collisions

joaotgouveia:
Changes how operators that are spelled using `<` or `>` are printed. Printing `<` or `>` collided with the search performed by `Mapper::matchTemplate`.


cpp2rust/converter/mapper.cpp
```diff
@@ -922,8 +923,30 @@ std::string ToString(const clang::NamedDecl *decl) {
...
+      os << "shr";
+      break;
+    case clang::OverloadedOperatorKind::OO_LessLessEqual:
+      os << "shleq";
+      break;
+    case clang::OverloadedOperatorKind::OO_GreaterGreaterEqual:
+      os << "shreq";
+      break;
+    default:
+      std::unreachable();
```
lucic71: Use assert. The rest of the PR looks good

PR #315, opened 2026-08-19
Avoid generating empty `thread_local`s

FranciscoTGouveia:
A new `IsConvertableGlobalVarDecl` function was introduced to determine whether the global variable declaration would emit any code.
This was needed because empty thread locals were previously being generated (i.e., `thread_local!()`).

### DIFF
--- a/cpp2rust/converter/converter.cpp
+++ b/cpp2rust/converter/converter.cpp
@@ -457,6 +457,17 @@ void Converter::ConvertVaListVarDecl(clang::VarDecl *decl) {
   StrCat(keyword_mut_, GetNamedDeclAsString(decl), token::kColon, "VaList");
 }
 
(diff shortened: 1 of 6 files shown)


nunoplopes:
This patch seems weird. Isn't the code exactly the same before/after, just less efficient?

FranciscoTGouveia:
Before, we would have:
```
thread_local!(pub static x_0: Value<i32> = <Value<i32>>::default());
thread_local!();
```
Whereas now, we just have:
```
thread_local!(pub static x_0: Value<i32> = <Value<i32>>::default());
```

This has no performance implications, but it avoids generating meaningless statements.

nunoplopes:
I mean performance of cpp2rust. This patch does 2 hash lookups instead of 1.

Issue #220, opened 2026-07-01
Incorrect substitutions for rules using `Ptr` parameters


joaotgouveia:
Using the following rules for `std::vector`:
```cpp
// src.cpp
template <typename T1> using t1 = std::vector<T1>;

template <typename T1> std::vector<T1> f1(const std::vector<T1> &a0) {
  return std::vector<T1>(a0);
}

template <typename T1> std::vector<T1> f2(std::vector<T1> &&a0) {
  return std::vector<T1>(std::move(a0));
}

template <typename T1>
std::vector<T1> &f3(std::vector<T1> &a0, std::vector<T1> &&a1) {
  return a0.operator=(std::move(a1));
}
```
```rust
// tgt_refcount.rs
fn t1<T1>() -> Vec<T1> {
    Vec::new()
}

fn f1<T1: Clone>(a0: Ptr<Vec<T1>>) -> Vec<T1> {
    a0.to_strong().borrow().clone()
}

fn f2<T1>(a0: Ptr<Vec<T1>>) -> Vec<T1> {
    std::mem::take(&mut *a0.to_strong().borrow_mut())
}

fn f3<T1: ByteRepr>(a0: Ptr<Vec<T1>>, a1: Ptr<Vec<T1>>) -> Ptr<Vec<T1>> {
    let moved = a1.with_mut(|v| std::mem::take(v));
    a0.with_mut(|dst| *dst = moved);
    a0
}
```

The following translation is generated:
```cpp
int main() {
  std::vector<int> vec1;
  std::vector<int> vec2(std::move(vec1));
  std::vector<int> vec3(vec2);
  vec3 = std::move(vec2);

  return 0;
}
```
```rust
fn main_0() -> i32 {
    let vec1: Value<Vec<i32>> = Rc::new(RefCell::new(Vec::new()));
    let vec2: Value<Vec<i32>> = Rc::new(RefCell::new(std::mem::take(
        &mut *(*vec1.borrow_mut()).to_strong().borrow_mut(), // (*vec1.borrow_mut()) should be (vec1.as_pointer() as Ptr<...>)
    )));
    let vec3: Value<Vec<i32>> =
        Rc::new(RefCell::new(vec2.as_pointer().to_strong().borrow().clone())); // vec2.as_pointer should be (vec2.as_pointer() as Ptr<...>)
    {
[...]

PR #318, opened 2026-08-26
Fix rvalue reference translation

joaotgouveia:
Translates rvalue references to `Ptr<...>`.


joaotgouveia:
The only thing missing to completely fix #220 is changing this condition:
https://github.com/Cpp2Rust/cpp2rust/blob/2cfd406ad421fd7ad3b861a4a03423d725011efb/cpp2rust/converter/converter.cpp#L4206
To something like this:
```cpp
if (ph_ctx.needs_pointer_receiver() && ph_ctx.needs_ptr_wrap()) {
```
However, this change introduces an error in the Brunslli translation due to #320 .

cpp2rust/converter/models/converter_refcount.cpp
```diff
@@ -364,11 +368,14 @@ ConverterRefCount::MaterializeTemp(const std::string &binding_name,
   auto pointee = param_type.getNonReferenceType();
   auto value = ConvertRValue(expr, pointee);
   auto type_str = ToStringBase(pointee);
-  std::string binding =
-      std::format("let {} : Value < {} > = Rc::new(RefCell::new( {} )) ;",
-                  binding_name, type_str, value);
-  std::string ref = std::format("{}.as_pointer()", binding_name);
-  return {binding, ref};
+  auto decl = in_const_initializer_ ? std::string("pub ") + keyword::kStatic
```
nunoplopes: `const char *decl = is_const_initializer ? "pub static" : ...`

Otherwise kittens die with a heartattack.

tests/unit/rvalue_ref_goto.cpp
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2022-present INESC-ID.
```
lucic71: Tests don't need the banner

tests/unit/rvalue_ref_general.cpp
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2022-present INESC-ID.
+// Distributed under the MIT license that can be found in the LICENSE file.
+
+#include <cassert>
+#include <utility>
+
+int main() {
+  int i1 = 3;
+  int i2 = std::move(i1);
```
lucic71, line 6: Add also a test where we have a value of type && and the initializer is std::move

lucic71: Also add a test where a struct is moved

tests/unit/out/unsafe/struct_ctor.rs
```diff
@@ -35,7 +35,7 @@ pub fn main() {
     }
 }
 unsafe fn main_0() -> i32 {
-    let mut struct_with_ctor: StructWithCtor = StructWithCtor::StructWithCtor(1, 2);
+    let mut struct_with_ctor: StructWithCtor = StructWithCtor::StructWithCtor({ 1 }, { 2 });
```
lucic71, line 38: Why are the extra {} needed?

tests/unit/out/refcount/rvalue_ref_global.rs
```diff
@@ -0,0 +1,19 @@
...
+use libcc2rs::*;
+use std::cell::RefCell;
+use std::collections::BTreeMap;
+use std::io::prelude::*;
+use std::io::{Read, Seek, Write};
+use std::os::fd::AsFd;
+use std::rc::{Rc, Weak};
+thread_local!(
+    pub static __tmp_0: Value<i32> = Rc::new(RefCell::new(5));
+    pub static g_0: Ptr<i32> = __tmp_0.with(Value::as_pointer);
```
lucic71, line 11: Isn't this pointer dangling because it's leaking the address of a block-local variable, ie __tmp_0?

tests/unit/out/refcount/lvalue_ref_goto.rs
```diff
@@ -0,0 +1,26 @@
...
+use std::io::prelude::*;
+use std::io::{Read, Seek, Write};
+use std::os::fd::AsFd;
+use std::rc::{Rc, Weak};
+pub fn main() {
+    std::process::exit(main_0());
+}
+fn main_0() -> i32 {
+    let __tmp_0: Value<i32> = Rc::new(RefCell::new(5));
+    let mut r: Ptr<i32> = <Ptr<i32>>::default();
```
lucic71, line 14: We should not track the mutability using mut in refcount. Can't we wrap this in a Value?

tests/ub/out/refcount/ret_ref_to_local.rs
```diff
@@ -0,0 +1,19 @@
...
+use libcc2rs::*;
+use std::cell::RefCell;
+use std::collections::BTreeMap;
+use std::io::prelude::*;
+use std::io::{Read, Seek, Write};
+use std::os::fd::AsFd;
+use std::rc::{Rc, Weak};
+pub fn foo_0() -> Ptr<i32> {
+    let __tmp_0: Value<i32> = Rc::new(RefCell::new(5));
+    return __tmp_0.as_pointer();
```
lucic71, line 11: Doesn't this dangle?

lucic71:
Also can we have a test where we move a struct?

tests/unit/out/refcount/rvalue_ref_global.rs
joaotgouveia: Per the docs: "The macro wraps any number of static declarations and makes them thread local. Publicity and attributes for each static are allowed."

Additionally, this test passes, which serves as further confirmation that the `__tmp_0` variable is not dropped.

tests/ub/out/refcount/ret_ref_to_local.rs
joaotgouveia: Yes, this test is regarding UB.

tests/unit/out/refcount/lvalue_ref_goto.rs
joaotgouveia: With the current approach, the only situation in which we need a `mut` variable in refcount is when we hoist a reference when translating `goto`. This is because the `Ptr` is first declared, and then assigned to. This can probably be wrapped in a `Value` instead of adding the extra `mut`, but I'd say the current approach seems simpler. Considering that the extra `Value` should, in theory, be optimized out anyway, I think that it may be unnecessary to wrap this particular case in a `Value`.

tests/unit/out/unsafe/struct_ctor.rs
joaotgouveia: They are not needed on that particular test, the braces were introduced by the changes to `Converter::ConvertCXXConstructExprArgs`. These changes ensure the test `tests/ub/ctor_ref_member.cpp` is correctly translated.

PR #332, opened 2026-09-06
Translate in-class initializers in Default trait

lucic71:


tests/unit/in_class_field_initializer.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+struct Inner {
+  int x = 3;
+  int y = 4;
+};
+
+struct S {
+  int a = 1;
+  char b = 2;
+  Inner c = {};
```
nunoplopes, line 12: Pls add another test:
```
Inner d;
```

should be initialized like `c`

lucic71: Done

tests/unit/out/unsafe/in_class_field_initializer.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub b: libc::c_char,
+    pub c: Inner,
+    pub d: Inner,
+}
+impl Default for S {
+    fn default() -> Self {
+        S {
+            a: 1,
+            b: (2 as libc::c_char),
+            c: Inner { x: 3, y: 4 },
```
nunoplopes, line 33: c should be initialized like d

lucic71: Even if Inner is defined as:

```cpp
struct Inner {
  int x; // was int x = 3;
  int y = 4;
}
```

?

nunoplopes: `Inner x{}` would initialize 'x' to zero, While 'Inner x' doesn't. But we probably need to default initialize everything anyway.

lucic71: I converted all struct defaults as default structure literals `Inner { x: 0, y : 4 }` instead of `<Inner>::default()`

nunoplopes: I meant the other way around: we should call Inner::default rather than duplicating it's body here.

lucic71: Done

PR #335, opened 2026-09-07
Translate operators

lucic71:
This PR adds support for the following operators:
* arithmetic: `+, -, unary +, unary -, *, /, %, ++ prefix and postfix, -- prefix and postfix`
* relational: `==, !=, >, <, >=, <=, <=>`
* logical: `!, &&, ||`
* bitwise: `~a, &, |, ^, <<, >>`
* assignment: `=, +=, -=, *=, /=, %=, &=, |=, ^=, <<=, >>=`
* member and pointer: `[] and static [], *, &, ->`
* other: `() and static (), , (comma), user defined conversion`

Not supported:
* `<<` for streams
* `a->*b` pointer to member
* `"a"_b` user defined literal
* `new`, `delete`

In Rust, operators are translated as `operator_call`, `operator_comma`, etc. Operators that are cv or ref-qualified contain that information in the name, for example `operator_call_const_volatile_rref`.

We support both method versions and free function versions of operators. Furthermore, we also support static `operator[]` and static `operator()`.

The following operators are used to generate Rust traits:
* `operator<` or `operator<=>` generate PartialOrd and Ord
* `operator==` or `operator<` or `operator<=>` are used for PartialEq and Eq
* the operators are implemented on `Ptr<Self>` while the traits are implemented on `Self`

I also added rules and tests for `operator<=>`. Since this is a C++20 feature, I had to change CMakeLists, rules/ and Cpp2RustTest.py in order to compile everything related to `operator<=>` with `-std=c++20`.


lucic71:
This PR is big because I added a lot of tests. Otherwise, the changes in cpp2rust are not that big

PR #338, opened 2026-09-09
Shadow `self` receiver with `this` pointer in unsafe

lucic71:
In unsafe, method receivers have type `&Self`. This is incompatible for pointer comparisons between `this` and other pointers, for example:

```cpp
bool copy_if_different_const(const S* other) {
  if (this == other) {
    return false;
  }
  ...
}
```

Translates to:

```rs
fn copy_if_different_const(&mut self, other: *const S) {
  if self == other { return false; }
  ...
```

In rust, comparing a mutable reference with a const pointer is not allowed. To fix this, I added a function preamble that creates a this pointer with mutability attached:

```rs
let this = self as *mut S
```

and the check becomes:

```rs
if this.cast_const() == other { return false; }
```

### DIFF
--- a/cpp2rust/converter/converter.cpp
+++ b/cpp2rust/converter/converter.cpp
@@ -2293,8 +2293,7 @@ bool Converter::VisitImplicitCastExpr(clang::ImplicitCastExpr *expr) {
     const char *suffix = nullptr;
     bool type_changed = false;
     if (expr->getType()->isPointerType() &&
-        sub_expr->getType()->isPointerType() &&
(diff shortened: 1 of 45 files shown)


nunoplopes:
Seems to go in the wrong direction. We want native Rust, not a copy of C++, so introducing a new variable called this isn't a good idea. You can cast on-demand; it's a free operation.

lucic71:
> Seems to go in the wrong direction. We want native Rust, not a copy of C++, so introducing a new variable called this isn't a good idea. You can cast on-demand; it's a free operation.

I did another PR that only convertrs self to pointer when needed

PR #342, opened 2026-09-09
Translate copy constructor and copy assignment

lucic71:
Copy constructors are translated in the `impl Self` block, they take no receiver and build the `Self` object using the provided argument.

If the copy constructor is not defaulted and not deleted, both unsafe and refcount generate it and Clone calls it.

```rs
impl Counted {
    // Copy constructor
    pub unsafe fn Counted_pconstCounted(o: *const Counted) -> Self {
        let mut this = Self { v: (*o).v };
        copies_0.prefix_inc();
        this
    }
}
impl Clone for Counted {
    // Clone implementation calls the copy constructor
    fn clone(&self) -> Self {
        unsafe { Counted::Counted_pconstCounted(self as *const Counted) }
    }
}
```

If the copy constructor is defaulted, unsafe continues to derive Clone and refcount continues to generate a deep copy:

```rs
fn clone(&self) -> Self {
  Self {
    field: Rc::new(RefCell::new(field.clone()));
    ...
  }
}
```

Copy assignment is translated using `operator_assign`:

```rs
impl PartialImpl for Ptr<Partial> {
    fn operator_assign(&self, o: Ptr<Partial>) -> Ptr<Partial> {
        if ((*self) == (o)) {
            return (*self).clone();
        }
        ...
        return (*self).clone();
    }
}
```

And users of copy assignment are translated as `PartialImpl::operator_assign(&a.as_pointer(), b.as_pointer())`


lucic71:
Depends on #341

PR #343, opened 2026-09-09
Translate move constructor and move assignment

lucic71:
Move constructor is translated as a normal constructor:

```rs
impl MoveOnly {
    pub unsafe fn MoveOnly_pmutMoveOnly(o: *mut MoveOnly) -> Self {
        let mut this = Self { v: (*o).v };
        (*o).v = 0;
        this
    }
}
```

Move assignment is translated using `operator_assign`:

```rs
impl MoveOnly {
    pub unsafe fn operator_assign_pmutMoveOnly(&mut self, o: *mut MoveOnly) -> *mut MoveOnly {
        if ((self as *mut MoveOnly) == (o)) {
            return &mut (*(self as *mut MoveOnly)) as *mut MoveOnly;
        }
        self.v = (*o).v;
        (*o).v = 0;
        return &mut (*(self as *mut MoveOnly)) as *mut MoveOnly;
    }
}
```

The combination defaulted move constructor and non-default copy constructor is not supported yet. `std::move` is translated as a clone() currently. It assumes that a clone is bitwise, i.e. the copy constructor is implicit. If it's not, the `std::move` would call the user defined copy constructor which is wrong. I will work on this in a future PR.


lucic71:
Depends on #342

PR #345, opened 2026-09-09
Translate if constexpr

lucic71:


cpp2rust/converter/converter.cpp
```diff
@@ -1254,6 +1254,14 @@ void Converter::ConvertCondition(clang::Expr *cond) {
 }
 
 bool Converter::VisitIfStmt(clang::IfStmt *stmt) {
+  if (stmt->isConstexpr()) {
+    if (auto taken = stmt->getNondiscardedCase(ctx_)) {
```
nunoplopes, line 1258: well, it would be good to keep all branches for documentation purposes. Constant folding here is not what a real project wants in general.

lucic71: I preserved the untaken branches where possible. In a template function, Clang is not required to keep the untaken branches, so it replaces them with NullStmt, so there is nothing to preserve there.

In non-template functions, Clang preserves the untaken branches. I also kept them in the generated code.

PR #347, opened 2026-09-09
Translate concepts and requires expressions

lucic71:
They are evaluated by clang. Only the result of the evaluation is translated.

Furthermore, static_assert is no-op because it's evaluated at compile time. Same for concept declaration.

### DIFF
--- a/cpp2rust/converter/converter.cpp
+++ b/cpp2rust/converter/converter.cpp
@@ -3421,6 +3421,13 @@ bool Converter::VisitConceptSpecializationExpr(
   return false;
 }
 
+bool Converter::VisitRequiresExpr(clang::RequiresExpr *expr) {
+  assert(!expr->isValueDependent());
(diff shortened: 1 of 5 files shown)


nunoplopes:
It would be nice to translate these statements rather than constant fold at compile time.

lucic71:
> It would be nice to translate these statements rather than constant fold at compile time.

I think i can keep static_asserts, concepts and requires would be too difficult to keep now.

lucic71:
> It would be nice to translate these statements rather than constant fold at compile time.

Done, I translate static_assert as const assert in Rust.

For the things that can be const evaluated in rust I keep the expression inside static_assert, e.g. `const _: () = assert!(size_of(int) == 4)`.

For the things that cannot be const-evaluated (non-const Rust functions) or for the things that don't have a direct translation to Rust (concepts and requires) I write: `const _: () = assert!(true, "Small<int>")` to keep track of the C++ expression that generated the code.

PR #353, opened 2026-09-10
Mangle overloaded methods

lucic71:


cpp2rust/converter/mapper.cpp
nunoplopes: this loop can be removed

PR #359, opened 2026-09-13
Bump to LLVM 23

lucic71:


lucic71:
@nunoplopes this fixes the problems in https://github.com/Cpp2Rust/cpp2rust/pull/358. No new rule is needed. There was a mismatch between the toolchain that cpp-rule-preprocessor uses and the one that cpp2rust uses.

Issue #363, opened 2026-09-14
Freshness is not tracked correctly


lucic71:
Not all producers of fresh values set `computed_expr_type_`. This results in unneeded clones. 

PendingDeref should also track freshness separately.

PR #364, opened 2026-09-14
Translate user defined moves

lucic71:
This PR translates move constructors (defaulted or implicit or user written) as ordinary constructors: `fn S_pmutS(_a0: *mut S) -> Self`. Defaulted and implicit ones are first synthesized by Sema.

Some notes about the implementation:

1. This below check is related to https://github.com/Cpp2Rust/cpp2rust/issues/363. It will disappear after the issue is solved.

```cpp
  // TODO: set freshness correctly to avoid stale computed_expr_type_
  if (expr->isGLValue()) {
    SetValueFreshness(expr->getType());
  }
```

2. Rules that move always translate to `std::mem::take` which is wrong if the argument of the rule has a user defined move constructor. I will solve this in a future PR (#365) where I will rewrite `std::mem::take` to use the move constructor if the argument has one.


lucic71:
Depends on #362

lucic71:
I will close this and split this PR into multiple ones. The converter is fine, but the tests fail because of library differences.

The STL between my local machine and ubuntu/macos is different. The STL in the runners forces the instantiation of move constructors that my local library does not force. Because of this, in the runners there are extra move constructors that are not present in the generated code on my machine.

For example:

```cpp
struct S { int a; }

std::vector<S> v;
// in the runner's library, this forces a move constructor on S that my library does not have
v.erase(...);
```

Even if I update my library to be on par with ubuntu, the macos will still fail.

So I will translate all move constructors from all user defined structs, even if they are not used in the code.

PR #365, opened 2026-09-14
Rewrite `std::mem::take` in rules for records with user-defined move constructor

lucic71:
Calling `std::mem::take` moves the objects and leaves a default constructed value in place. This is wrong for records with user defined move constructors because it drops the side effects inside the user defined move constructor.

This PR rewrites `std::mem::take` from rules as follows:
* record has move constructor => rewrite `std::mem::take` as a call to the move constructor
* record does not have move constructor and is copyable => convert as rvalue (without clone)
* record does not have move constructor and is not copyable => convert as fresh rvalue (with clone)
* record is STL type => use the rule for STL move constructor
* temporary value => convert as rvalue


lucic71:
Depends on #364

PR #367, opened 2026-09-14
Fix #114: Rust 2024 compatibility

### LINKED ISSUE #114: Rust 2024 compatibility
Rust 2024 doesn't allow references to static mut variables. Some tests fail:
```
********************
FAIL: cpp2rust :: unit/bool_condition_logical.cpp/unsafe (24 of 634)
******************** TEST 'cpp2rust :: unit/bool_condition_logical.cpp/unsafe' FAILED ********************
rustc failed
error: creating a mutable reference to mutable static
  --> main.rs:28:5
   |
28 |     side_effect.prefix_inc();
   |     ^^^^^^^^^^^^^^^^^^^^^^^^ mutable reference to mutable static
   |
   = note: mutable references to mutable statics are dangerous; it's undefined behavior if any other pointer to the static is used or if any other reference is created for the static while the mutable reference lives
   = note: for more information, see <https://doc.rust-lang.org/edition-guide/rust-2024/static-mut-references.html>
   = note: `#[deny(static_mut_refs)]` (part of `#[deny(rust_2024_compatibility)]`) on by default

error: aborting due to 1 previous error
```

After fixing this, remove `-W static_mut_refs` from the `Cpp2RustTest.py` test driver.

nunoplopes:


lucic71:
I think this is already solved in #366

cpp2rust/converter/converter.cpp
```diff
@@ -30,6 +30,20 @@ std::unordered_set<std::string> Converter::abstract_structs_;
 Converter::RecordIndex Converter::record_decls_;
 std::map<std::string, Converter::MethodsOnPtr> Converter::methods_on_ptr_;
 
+// `static mut` globals cannot have a `&`/`&mut` reference formed directly to
+// them in Rust 2024 edition. Wrapping the variable in a `&raw mut` pointer
+// followed by a deref sidesteps the restriction.
+// Doesn't apply when a rule replaced the expression with custom Rust code.
+static std::string WrapMutableStaticPlace(const clang::Decl *decl0,
```
lucic71, line 37: This can be inlined to avoid passing the string then returning it

PR #369, opened 2026-09-14
Translate user defined moves

lucic71:
This PR translates move constructors as ordinary constructors: `fn S_pmutS(_a0: *mut S) -> Self`.

The rules for translating and using the move constructors are:

|No | Move Constructor (MC) | Copy Constructor | Result |
| --- | --- | --- | --- |
| 1 | user-written | implicit/defaulted/deleted | translate MC and always use it |
| 2 | user-written | user-written | translate MC and always use it |
| 3 | implicit/defaulted | implicit/defaulted | don't translate MC and use `.clone()` on move |
| 4 | implicit/defautled | user-written/deleted | translate MC and always use it |

For 1 and 2, we always use the translated move constructor because it might contain custom logic that needs to run on every move. For 3 we use `.clone()` because it has the same effect as a move would have. For 4 `.clone()` is not available (deleted) or might contain custom logic (user-written) so we synthesize an implicit/defaulted move constructor and use it.

`std::move` becomes a transparent construct, i.e. no move decision is taken at the time of the cast. It's only used by clang to choose the right constructor. Moves happen on CXXConstructExpr.

Each STL type has move construction rules. Rules that move using `std::mem::take` have their bodies rewritten as follows:
* record has move constructor => rewrite `std::mem::take` as a call to the move constructor
* record does not have move constructor and is copyable => convert as rvalue (without clone)
* record does not have move constructor and is not copyable => convert as fresh rvalue (with clone)
* record is STL type => use the rule for STL move constructor
* temporary value => convert as rvalue


tests/unit/copy_move_defaulted.cpp
```diff
@@ -101,5 +127,40 @@ int main() {
...
+  assert(o2.n == 5 && o2.data.size() == 1 && o2.arr[1] == 6 && *o2.p == 7);
+  assert(o1.data.empty() && o1.p.get() == nullptr);
+  Owner o3;
+  o3 = std::move(o2);
+  assert(o3.n == 5 && o3.data[0] == 5 && o3.arr[0] == 5 && *o3.p == 7);
+  assert(o2.data.empty() && o2.p.get() == nullptr);
+
+  Holder h1(4);
+  h1.p.reset(new int(9));
+  Holder h2 = std::move(h1);
```
nunoplopes, line 159: the translation of this is calling the copy constructor rather than move, I think.

lucic71: It's the move constructor. Copy constructor is deleted because Holder contains unique_ptr which is not copy

nunoplopes:
This strategy works while fields are boxed in Values, but otherwise we need move constructors to call `std::mem::take` to move objects and destroy (default initialize) the old one, otherwise we break the translation of classes for example that do reference counting.
I would also like to see more meaningful names in functions. It's very hard to follow the generated code.

PR #370, opened 2026-09-14
Add codegen/types docs

lucic71:
This doesn't have all the info updated. For example enums are no more transalted as enum in Rust, I plan to update that in a future PR. For the moment I would like to get the core of codegen/types merged, then I will make the appropiate edits.


nunoplopes:
mermaid.js should be kept outside of the repo.  The github actions can download the thing from e.g. `https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.min.js`

lucic71:
> mermaid.js should be kept outside of the repo. The github actions can download the thing from e.g. `https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.min.js`

Done

PR #372, opened 2026-09-16
Make rule-preprocessor process a single directory at a time

joaotgouveia:
Changes `rule-preprocessor` to process a single directory per invocation, making it possible to parallelize. This matches what is already the case for `cpp-rule-preprocessor`. Additionally, this drops `modules.rs` from source control.


CMakeLists.txt
```diff
@@ -141,18 +141,46 @@ set(RULES_IR_DIR "${CMAKE_BINARY_DIR}/rules")
...
+     ${PROJECT_SOURCE_DIR}/rule-preprocessor/src/*.rs)
+
+execute_process(
+  COMMAND ${CMAKE_COMMAND} -E env RUSTUP_TOOLCHAIN=${RUST_NIGHTLY_VERSION}
+          rustc --print sysroot
+  OUTPUT_VARIABLE NIGHTLY_SYSROOT
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+
+add_custom_command(
```
nunoplopes, line 157: this should be split in two rules, each output its own binary. Then there's no need for this `PREPROCESSOR_STAMP`

joaotgouveia: I'll split building the `rules` crate and the `rule-preprocessor`, keeping both commands together is not necessary since they are independent. However, preprocessing rust rules depends on both commands, one ensures the binary itself is build, and the other ensures that the `rules` crate dependencies are built.

Since preprocessing rust rules depends on the built dependencies, and not on the `RUSTLS_HEADER` the `rules` crate command outputs, won't it be clearer to keep the `PREPROCESSOR_STAMP` as an output of this command as well?

PR #410, opened 2026-09-20
Remove Ptr::to_strong() and add Ptr::decay()

nunoplopes:


cpp2rust/converter/models/converter_refcount.cpp
```diff
@@ -321,18 +336,43 @@ std::string ConverterRefCount::ConvertFreshLValue(clang::Expr *expr) {
...
-  auto str = ConvertObject(expr);
+std::string
+ConverterRefCount::ConvertFreshObject(clang::Expr *expr,
+                                      std::string_view target_ptr_type) {
+  auto shape = ObjectShape::Whole;
+  if (!target_ptr_type.empty()) {
+    auto type = expr->getType().getNonReferenceType();
+    auto pointee = type->isPointerType() ? type->getPointeeType() : type;
+    if (IsBoxedType(pointee) || pointee->isArrayType()) {
+      auto normalize = [](std::string s) {
```
lucic71, line 363: This is a hack. The correct solution is to have more type information in Rules IR so that we can write queries like `T.is_pointer() && T.pointee() == X`.

cpp2rust/converter/models/converter_refcount.h
```diff
@@ -369,6 +375,7 @@ class ConverterRefCount final : public Converter {
   std::string BoxValue(std::string &&str) const;
 
   std::vector<ConversionKind> conversion_kind_;
+  ObjectShape object_shape_ = ObjectShape::Whole;
```
lucic71, line 378: Why can't ExprKind be extended instead of adding a new enum?

